PR #1, opened 2020-03-27
JAVA-2682: Create a Quarkus Cassandra extension

tomekl007:


adutra:
General remark: I'm surprised by the general layout. 

1. Why do we need to inherit from JBoss parent POM?
1. Why do we need a build-parent module, couldn't it be merged into the root POM?
1. If poms/bom is the runtime BOM, can we name it `bom-runtime` for clarity?
1. The runtime module is called `cassandra-quarkus-client`, shouldn't we rename it `cassandra-quarkus-client-runtime` for symmetry with `cassandra-quarkus-client-deployment`?
1. Why do we need the intermediary `cassandra-quarkus-extensions` POM?
1. Why is `cassandra-quarkus-integration-tests` a POM module with a child JAR module?
1. Some properties are declared and only used in one place, I don't find it particularly useful and creates a level of indirection that makes it hard to understand the POMs.

I haven't looked more in detail but it seems that the general structure of the project could be much simpler.

poms/bom/pom.xml
```diff
@@ -0,0 +1,149 @@
...
+
+            <!-- Cassandra Quarkus -->
+            <dependency>
+                <groupId>com.datastax.oss.quarkus</groupId>
+                <artifactId>cassandra-quarkus-client</artifactId>
+                <version>${cassandra-quarkus.version}</version>
+            </dependency>
+
+            <!-- Other third party dependencies here -->
+            <dependency>
```
adutra: Quarkus uses BOMs a lot, why don't we use the driver BOM as well instead of declaring the driver dependencies one by one?

poms/bom/pom.xml
```diff
@@ -0,0 +1,149 @@
...
+                <groupId>com.datastax.oss</groupId>
+                <artifactId>java-driver-mapper-runtime</artifactId>
+                <version>${datastax-java-driver.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>com.datastax.oss</groupId>
+                <artifactId>java-driver-mapper-processor</artifactId>
+                <version>${datastax-java-driver.version}</version>
+            </dependency>
+            <dependency>
```
adutra: Compression libraries should be optional, no?

extensions/cassandra-client/deployment/pom.xml
```diff
@@ -0,0 +1,114 @@
...
+                <groupId>com.datastax.oss.quarkus</groupId>
+                <artifactId>cassandra-quarkus-bom-deployment</artifactId>
+                <version>${project.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
```
adutra: The XML files are not well formatted, please add the XML format plugin that we use in the driver codebase.

poms/bom/pom.xml
tomekl007: yes, I've created a separate task for it:
https://datastax-oss.atlassian.net/browse/JAVA-2707

tomekl007:
> General remark: I'm surprised by the general layout.
> 
> 1. Why do we need to inherit from JBoss parent POM?
> 2. Why do we need a build-parent module, couldn't it be merged into the root POM?
> 3. If poms/bom is the runtime BOM, can we name it `bom-runtime` for clarity?
> 4. The runtime module is called `cassandra-quarkus-client`, shouldn't we rename it `cassandra-quarkus-client-runtime` for symmetry with `cassandra-quarkus-client-deployment`?
> 5. Why do we need the intermediary `cassandra-quarkus-extensions` POM?
> 6. Why is `cassandra-quarkus-integration-tests` a POM module with a child JAR module?
> 7. Some properties are declared and only used in one place, I don't find it particularly useful and creates a level of indirection that makes it hard to understand the POMs.
> 
> I haven't looked more in detail but it seems that the general structure of the project could be much simpler.

Overall, I wanted to mirror the structure from:
- https://github.com/riptano/quarkus
(I was also using https://github.com/Naros/debezium-quarkus and https://github.com/apache/camel-quarkus as the points of reference)

1. `quarkus` and `extensions` are inheriting this `JBoss` parent so If we want to be compatible I think we should as well.
2. The same situation - the `quarkus` project has dedicated `build-parent` module (and both extensions)
3. It seems that in the quarkus, for the `runtime` the suffix is omitted. I think that it is because the deployment "contains" runtime so those dependencies are not only for runtime.
4. It is the same situation -  this is the naming pattern that I took from the `quarkus` repo.
5. I think that we should unify it with quarkus and other extensions - for all of those, there is intermediary `extensions` module, and inside of it there are extensions (as a separate module, in our case it is  `cassandra-quarkus-extensions`)
6. Same as above.
7. I can inline them, I don't have a preference here.

So I totally agree that it could be simplified, but If we want to be unified with `quarkus` repo, and other extensions, we should follow this structure. Please let me know what do you think.

poms/bom/pom.xml
tomekl007: We would still need to do exclusions, so we would need to redefine them. 
I don't know if this will be beneficial, wdyt?

olim7t: I don't think the two are incompatible: the BOM only sets versions, you still need to declare (version-less) dependencies to driver artifacts in the `<dependencies>` section; these can have exclusions.

olim7t:
> The runtime module is called `cassandra-quarkus-client`, shouldn't we rename it `cassandra-quarkus-client-runtime` for symmetry with `cassandra-quarkus-client-deployment`?

Omitting the suffix for the runtime module is a Quarkus naming convention, see the [extension guide](https://quarkus.io/guides/writing-extensions#maven-setup).

On that note, the Quarkus maven plugin has a `create-extension` goal to bootstrap a new extension, @tomekl007 did you use that to initialize the project? If not, we could at least check that the structure we use is consistent with what it would have generated.

tomekl007:
> > The runtime module is called `cassandra-quarkus-client`, shouldn't we rename it `cassandra-quarkus-client-runtime` for symmetry with `cassandra-quarkus-client-deployment`?
> 
> Omitting the suffix for the runtime module is a Quarkus naming convention, see the [extension guide](https://quarkus.io/guides/writing-extensions#maven-setup).
> 
> On that note, the Quarkus maven plugin has a `create-extension` goal to bootstrap a new extension, @tomekl007 did you use that to initialize the project? If not, we could at least check that the structure we use is consistent with what it would have generated.
> 
> EDIT - output of the Maven plugin (but I think it's intended for a project that already has the `extensions/` module):
> 
> ```
> $ mvn io.quarkus:quarkus-maven-plugin:1.3.1.Final:create-extension -N \
>     -Dquarkus.artifactIdBase=cassandra-client \
>     -Dquarkus.groupId=com.datastax.oss \
>     -Dquarkus.nameBase="Cassandra client" \
>     -Dversion=1.0.0-SNAPSHOT
> $ find cassandra-client -type f
> cassandra-client/runtime/pom.xml
> cassandra-client/pom.xml
> cassandra-client/deployment/pom.xml
> cassandra-client/deployment/src/main/java/com/datastax/oss/cassandra/client/deployment/CassandraClientProcessor.java
> ```

Yes, I was using that, but exactly what you wrote " I think it's intended for a project that already has the extensions/ module", also this is intended only for extensions that are in the quarkus repo and rely on the structure that it's in that repo, so we cannot use it for our case.

tomekl007:
@adutra 
> General remark: I'm surprised by the general layout.
> 
> 1. Why do we need to inherit from JBoss parent POM?
> 2. Why do we need a build-parent module, couldn't it be merged into the root POM?
> 3. If poms/bom is the runtime BOM, can we name it `bom-runtime` for clarity?
> 4. The runtime module is called `cassandra-quarkus-client`, shouldn't we rename it `cassandra-quarkus-client-runtime` for symmetry with `cassandra-quarkus-client-deployment`?
> 5. Why do we need the intermediary `cassandra-quarkus-extensions` POM?
> 6. Why is `cassandra-quarkus-integration-tests` a POM module with a child JAR module?
> 7. Some properties are declared and only used in one place, I don't find it particularly useful and creates a level of indirection that makes it hard to understand the POMs.
> 
> I haven't looked more in detail but it seems that the general structure of the project could be much simpler.

I've simplified the project structure:

1. I change it to inherit `<artifactId>quarkus-parent</artifactId>` because this is recommended on the 
https://quarkus.io/guides/writing-extensions#maven-setup
2. I've removed `build-parent` and merged it with the root POM.
3. Naming convention https://quarkus.io/guides/writing-extensions#maven-setup
4. Naming convention
5. I've flattened it
6. I've flattened it.
7. I've inlined properties that are used only one time.

poms/bom/pom.xml
tomekl007: I've changed the pom to use the `datastax-java-driver` bom. If you are definition exclusions, the version definition needs to be repeated (but this is only one time)

pom.xml
```diff
@@ -0,0 +1,124 @@
...
+              <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+            </systemProperties>
+          </configuration>
+        </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-failsafe-plugin</artifactId>
+          <version>${maven-surefire-plugin.version}</version>
+        </plugin>
+        <plugin>
```
adutra, line 61: What is this plugin used for?

pom.xml
```diff
@@ -0,0 +1,124 @@
...
+              </goals>
+            </execution>
+          </executions>
+          <configuration>
+            <excludes>
+              <exclude>.idea/**</exclude>
+              <exclude>**/target/**</exclude>
+              <exclude>**/dependency-reduced-pom.xml</exclude>
+              <exclude>**/.flattened-pom.xml</exclude>
+              <exclude>docs/**</exclude>
```
adutra, line 101: This exclusion seems unused.

poms/bom/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+    <groupId>com.datastax.oss.quarkus</groupId>
+    <artifactId>cassandra-quarkus-parent</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
+    <relativePath>../../pom.xml</relativePath>
+  </parent>
+  <artifactId>cassandra-quarkus-bom</artifactId>
+  <packaging>pom</packaging>
+  <name>Cassandra Quarkus :: BOM :: Runtime</name>
+  <properties>
+    <cassandra-quarkus.version>1.0.0-SNAPSHOT</cassandra-quarkus.version>
```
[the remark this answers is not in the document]
tomekl007: yes, you are right. Done

poms/bom-deployment/pom.xml
```diff
@@ -0,0 +1,41 @@
...
+    <groupId>com.datastax.oss.quarkus</groupId>
+    <artifactId>cassandra-quarkus-parent</artifactId>
+    <version>1.0.0-SNAPSHOT</version>
+    <relativePath>../../pom.xml</relativePath>
+  </parent>
+  <artifactId>cassandra-quarkus-bom-deployment</artifactId>
+  <packaging>pom</packaging>
+  <name>Cassandra Quarkus :: BOM :: Deployment</name>
+  <properties>
+    <cassandra-quarkus.version>1.0.0-SNAPSHOT</cassandra-quarkus.version>
```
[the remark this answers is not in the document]
tomekl007: yes, you are right. Done

poms/bom/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+    <datastax-java-driver.version>4.5.1</datastax-java-driver.version>
+  </properties>
+  <dependencyManagement>
```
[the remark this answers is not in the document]
tomekl007: 1. Ok, I moved it to the runtime module `pom.xml`.
2. You are right - I've removed the definition of dependencies that are included in the `quarkus-parent` bom
3. I was also thinking that it may be a good idea because it will not cause a clash of dependencies if some module is not using the test dependency. Done.

pom.xml
tomekl007: copy-paste, sorry. Done

pom.xml
tomekl007: For running integration tests in the integration-tests module:
`mvn integration-test`

cassandra-extension/deployment/pom.xml
```diff
@@ -0,0 +1,123 @@
...
+      <artifactId>quarkus-core-deployment</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-smallrye-health-spi</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-smallrye-health</artifactId>
+      <scope>test</scope>
```
adutra, line 51: Are we sure we want this in test scope?

tomekl007: yes, you are right. We should convert it into `optional`. Done

integration-tests/src/main/java/com/datastax/oss/quarkus/CassandraNameConverterEndpoint.java
```diff
@@ -0,0 +1,45 @@
...
+import javax.inject.Inject;
+import javax.ws.rs.GET;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+@Path("/cassandra-name-converter")
+public class CassandraNameConverterEndpoint {
```
adutra, line 28: Would it be possible to simplify the tests to use just one endpoint, one service and one DAO + name converter? I'm having a hard time trying to follow what the tests do :-)

pom.xml
adutra: I was referring to the exec plugin, I removed it already.

cassandra-extension/runtime/src/main/java/com/datastax/oss/quarkus/runtime/AbstractCassandraClientProducer.java
```diff
@@ -0,0 +1,121 @@
...
+  public void setCassandraClientConfig(CassandraClientConfig config) {
+    this.config = config;
+  }
+
+  private ProgrammaticDriverConfigLoaderBuilder createDriverConfigLoader() {
+    return new DefaultProgrammaticDriverConfigLoaderBuilder(
+        () -> {
+          ConfigFactory.invalidateCaches();
+          return ConfigFactory.defaultOverrides()
+              // todo do we need to load both conf and json?
```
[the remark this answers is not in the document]
adutra: I removed the TODO, hope you don't mind, I think we need to support both.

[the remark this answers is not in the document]
tomekl007: Yes, thank you.

integration-tests/src/main/java/com/datastax/oss/quarkus/CassandraNameConverterEndpoint.java
tomekl007: I was also having a problem understanding those quarkus tests mainly because of the indirection layer (rest API) but this is recommended. 
Regarding two endpoints, I would like to have that split because of the `/cassandra` is testing a code path that does not involve reflection, whereas `cassandra-name-converter` tests naming converter that uses reflection, see:
```
@RegisterForReflection
public class TestNameConverter implements NameConverter {
```
We may extend those tests in the future, so I think that split is good here. wdyt?

cassandra-extension/runtime/src/main/java/com/datastax/oss/quarkus/config/CassandraClientConnectionConfig.java
```diff
@@ -0,0 +1,40 @@
...
+import java.util.List;
+import java.util.Optional;
+
+@ConfigGroup
+public class CassandraClientConnectionConfig {
+  /**
+   * Contact-points used to connect to Cassandra. If not specified, it will connect to local-host.
+   */
+  @ConfigItem(name = "contact-points", defaultValue = "127.0.0.1:9042")
+  public List<String> contactPoints;
```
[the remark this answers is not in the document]
tomekl007: Good idea, but if we will use `InetSocketAddress` we would rely on the quarkus on the creation of address and resolved vs unresolved behavior. When using `String` for `contact-point` we can be more flexible, and the driver itself can create `InetSocketAdress` however it wants. So I don't know if we should change it, wdyt?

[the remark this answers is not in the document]
adutra: Good point, we don't know and don't control how addresses are created. Let's leave it as List<String> for now, maybe later we can explore the idea of custom converters; according to the doc, they can be registered by adding their fully qualified class name in file `META-INF/services/org.eclipse.microprofile.config.spi.Converter`.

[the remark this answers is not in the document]
tomekl007: yes, I saw that option. We can work on it in a separate ticket, but still, it seems that the converter logic will be duplication of the driver logic for `InetAddresses`.

cassandra-quarkus-extension/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
```diff
@@ -0,0 +1,171 @@
...
+    return new HealthBuildItem(
+        "com.datastax.oss.quarkus.runtime.health.CassandraHealthCheck",
+        buildTimeConfig.healthEnabled,
+        "cassandra");
+  }
+
+  @BuildStep
+  RuntimeInitializedClassBuildItem runtimeMetadataManager() {
+    return new RuntimeInitializedClassBuildItem(MetadataManager.class.getCanonicalName());
+  }
```
olim7t: Why do we need to do this for `MetadataManager` specifically? It's just a regular internal driver class, there are many others like it.

pom.xml
```diff
@@ -0,0 +1,124 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <parent>
+    <groupId>io.quarkus</groupId>
+    <artifactId>quarkus-parent</artifactId>
+    <version>1.3.0.Final</version>
```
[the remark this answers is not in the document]
tomekl007: I think that it is beneficial for clients of this extension, to use the BOM. The reason is that if we are using the BOM, we don't need to define the dependencies that are provided in the quarkus, for example:
https://github.com/datastax/cassandra-quarkus/pull/1/commits/a7edd3f2fc14adb55318b51322c618af29d78c13
Because we are using BOM, we don't need to manage those dependencies by ourselves. If there will be a new version of the quarkus, we would need only to increment this BOM version, and all dependant dependencies will be automatically updated.
Also, it will help us with keeping dependencies in sync with other extensions (please note that the majority of extensions are in the `quarkus` repo `extension` module and they do extend this BOM). wdyt?

cassandra-quarkus-extension/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
tomekl007: MetadataManager must be initialized at runtime because it uses Inet4Socket address that cannot be initialized at deployment time because of:
```
No instances of java.net.Inet4Address are allowed in the image heap as this class should be initialized at image runtime.
```
I think that this is the most narrow scope that we can make as the `RuntimeItem` but maybe we can make it even more narrow. I was trying to substitute `Inet4Address` but I was unable to do so.

cassandra-quarkus-extension/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
```diff
@@ -0,0 +1,171 @@
...
+
+  private String getCassandraClientProducerClassName() {
+    return AbstractCassandraClientProducer.class.getPackage().getName()
+        + "."
+        + "CassandraClientProducer";
+  }
+
+  private void createCassandraClientProducerBean(
+      BuildProducer<GeneratedBeanBuildItem> generatedBean,
+      String cassandraClientProducerClassName) {
```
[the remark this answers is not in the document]
tomekl007: Yes, I wanted to simplify that, but the code is required because of the initialization order.
Please note the usage of `CassandraClientRecorder recorder,`. It is creating a proxy class that extends this auto-generated code. I was unable to make it work the way you suggested.

integration-tests/src/main/java/com/datastax/oss/quarkus/CassandraNameConverterEndpoint.java
adutra: But in this case, isn't testing the code path with reflection enough?

tomekl007: When I was developing it there were some cases when the `testSaveAndRetrieveProduct` was passing and `shouldSaveAndRetrieveUsingCustomNameConverterThatUsesReflection` was not so I think that is a good reason to have them separately. I think that tests should verify the narrow scope of code and give more precise feedback. If we will keep only the 2nd test, we would not be verifying the 1st code path.

pom.xml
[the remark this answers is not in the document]
adutra: But precisely, our BOMs already include Quarkus BOM, see [here](https://github.com/datastax/cassandra-quarkus/blob/8b4c1149738c69e6a0c52be5934f16483b2c4799/poms/bom/pom.xml#L36-L42), which in turn inherit from `quarkus-parent`. So all Quarkus dependencies are correctly made available to users of these BOMs.
Our project itself does not need to inherit from that parent POM.
I think we can safely get rid of that inheritance. The only thing to add will be these two properties:

```
    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
```

cassandra-quarkus-extension/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
adutra: Would you mind explaining that in a comment? Otherwise, for reviewers it does look strange to see this component special-cased.

cassandra-extension/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
```diff
@@ -0,0 +1,181 @@
...
+import javax.enterprise.inject.Default;
+import javax.enterprise.inject.Produces;
+
+class CassandraClientProcessor {
+  public static final String CASSANDRA_CLIENT = "cassandra-client";
+
+  @BuildStep
+  List<ReflectiveClassBuildItem> registerForReflection() {
+    return Arrays.asList(
+        new ReflectiveClassBuildItem(true, true, ExponentialReconnectionPolicy.class.getName()),
```
adutra, line 67: If I understand correctly, this list contains the default implementations that the driver loads by reflection. But happens if the users configures a different class, for example `DcInferringLoadBalancingPolicy`?

pom.xml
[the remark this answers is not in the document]
tomekl007: you right, the import of BOM is enough. Done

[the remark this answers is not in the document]
tomekl007: yes, you are right. Import of the BOM should be enough. Done

cassandra-quarkus-extension/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
tomekl007: good idea - done

cassandra-extension/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
tomekl007: Yes, we need to add all of those, not only defaults - thanks for pointing that out.
I added all of those. One question regarding `LoadBalancingPolicies` - should we expose the two that are deprecated:
`DseDcInferringLoadBalancingPolicy` and `DseLoadBalancingPolicy`? I think we should not - we should discourage users to use it.

adutra:
To reviewers: I pushed a few commits:

1. Project structure was simplified a bit more, you may need to reimport the project in your IDE, my apologies for the trouble.
2. Integration tests now run in the right Maven phase.
3. If you want to run the native tests, do `mvn clean verify -Dnative`. Make sure you have `GRAALVM_HOME` set.

cassandra-extension/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
adutra: I agree to not expose the deprecated ones.

cassandra-quarkus-extension/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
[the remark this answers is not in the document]
absurdfarce: FWIW I got this to work (where "work" means "the tests, or at least the non-native ones but including the integration tests, are passing") via the following:

https://github.com/absurdfarce/cassandra-quarkus/pull/1

Obviously that impl is extremely basic and not very idiomatic... I was largely just trying to demonstrate the idea.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/AbstractCassandraClientProducer.java
```diff
@@ -0,0 +1,120 @@
...
+      ProgrammaticDriverConfigLoaderBuilder configLoaderBuilder,
+      CassandraClientConnectionConfig config) {
+    configLoaderBuilder.withStringList(DefaultDriverOption.CONTACT_POINTS, config.contactPoints);
+    configLoaderBuilder.withString(
+        DefaultDriverOption.LOAD_BALANCING_LOCAL_DATACENTER, config.localDatacenter);
+    config.requestTimeout.ifPresent(
+        v -> configLoaderBuilder.withDuration(DefaultDriverOption.REQUEST_TIMEOUT, v));
+    // force java clock because jnr does not work with quarkus
+    configLoaderBuilder.withBoolean(DefaultDriverOption.TIMESTAMP_GENERATOR_FORCE_JAVA_CLOCK, true);
+  }
```
absurdfarce, line 82: I believe there was some talk about this in the previous PR but this doesn't seem like a scalable configuration strategy at all.  We're forced into a double enumeration here; we have to specify the config props in one place and then remember to add them in another.  :unpleasant:

Probably something else we can address after we get this ticket in... but if there isn't a ticket for this already there probably should be.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/AbstractCassandraClientProducer.java
```diff
@@ -0,0 +1,120 @@
...
+
+  public CqlSession createCassandraClient(CassandraClientConfig config) {
+    ProgrammaticDriverConfigLoaderBuilder configLoaderBuilder = createDriverConfigLoader();
+    configureConnectionSettings(configLoaderBuilder, config.cassandraClientConnectionConfig);
+    CqlSessionBuilder builder = CqlSession.builder().withConfigLoader(configLoaderBuilder.build());
+    return builder.build();
+  }
+
+  private void configureConnectionSettings(
```
absurdfarce, line 72: Semi-nit: void methods make me cringe a bit.  Can we change this to something like withConnectionSettings() and have it return the PDCLB rather than accomplishing it's mission through side effects?

runtime/src/main/java/com/datastax/oss/quarkus/runtime/AbstractCassandraClientProducer.java
```diff
@@ -0,0 +1,120 @@
...
+              .withFallback(ConfigFactory.parseResources("application.conf"))
+              .withFallback(ConfigFactory.parseResources("application.json"))
+              .withFallback(ConfigFactory.defaultReference())
+              .resolve();
+        },
+        DefaultDriverConfigLoader.DEFAULT_ROOT_PATH) {
+      @NonNull
+      @Override
+      public DriverConfigLoader build() {
+        return new NonReloadableDriverConfigLoader(super.build());
```
absurdfarce, line 56: Apologies if I missed this somewhere.. is there a reason we need a non-reloadable ConfigLoader here?  We're declaring a new type to get that behaviour so I'm _assuming_ we do... but I'm not quite sure I understand the rationale.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/graal/NativeSubstitutions.java
```diff
@@ -0,0 +1,34 @@
...
+ * limitations under the License.
+ */
+package com.datastax.oss.quarkus.runtime.graal;
+
+import com.datastax.oss.driver.internal.core.os.Native;
+import com.oracle.svm.core.annotate.Substitute;
+import com.oracle.svm.core.annotate.TargetClass;
+
+@TargetClass(Native.class)
+final class NativeSubstitutions {
```
absurdfarce, line 23: I'm musing a bit here... do we want to include isCurrentTimeMicrosAvailable() here as well (with a constant return value of false) until we get the native stuff sorted out?  We're side-stepping the issue entirely now by forcing the use of a Java clock but my thought was that that's an indirect consequence of not having native support.

On the other hand that clock is the only consumer of this functionality... so I can see a counter-argument which says adding it here is just one more thing we have to unset/change when we want to confirm native functionality is fine.

I guess I could go either way here.

deployment/src/test/java/com/datastax/oss/quarkus/CassandraClientConfigTest.java
```diff
@@ -0,0 +1,76 @@
...
+
+    assertThat(profile.getDuration(REQUEST_TIMEOUT)).isEqualTo(Duration.of(20, ChronoUnit.SECONDS));
+  }
+
+  @Test
+  public void applicationJsonSettingsShouldHavePriorityOverReferenceConfFromDriver() {
+    DriverExecutionProfile profile = cqlSession.getContext().getConfig().getDefaultProfile();
+
+    assertThat(profile.getInt(REQUEST_PAGE_SIZE)).isEqualTo(1000);
+  }
```
absurdfarce, line 75: At some point it might be desirable to build the files we're adding to the archive as part of the test rather than storing them externally as resources.  That way we can have the logic for test data generation and validation both in the same place (this test) rather than split across two locations (this test and the resources).

Also a thing we can improve in the future.

integration-tests/src/main/java/com/datastax/oss/quarkus/dao/nameconverters/NameConverterEntity.java
```diff
@@ -0,0 +1,41 @@
...
+ */
+package com.datastax.oss.quarkus.dao.nameconverters;
+
+import com.datastax.oss.driver.api.mapper.annotations.Entity;
+import com.datastax.oss.driver.api.mapper.annotations.NamingStrategy;
+import com.datastax.oss.driver.api.mapper.annotations.PartitionKey;
+
+@Entity
+@NamingStrategy(customConverterClass = TestNameConverter.class)
+public class NameConverterEntity {
```
absurdfarce, line 24: In the spirit of facilitating comprehension it might be better to rename this entity "ReflectionEntity" or something along those lines to clearly indicate what makes it unique.  Same argument applies for the entire chain of services and endpoints which ultimately lead to this class.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/AbstractCassandraClientProducer.java
tomekl007: Yes, right now we decided to go with this approach:
- basic properties explicitly set using the quarkus config mechanism, all other driver settings could be passed using `application.conf` or `application.json`(see `CassandraClientConfigTest`).
There is also a task to improve it: https://datastax-oss.atlassian.net/browse/JAVA-2697

runtime/src/main/java/com/datastax/oss/quarkus/runtime/AbstractCassandraClientProducer.java
tomekl007: The plan is to not support reloading of config, we will want to push as much config options to deployment module as possible -  in such a case reloading will not be possible

runtime/src/main/java/com/datastax/oss/quarkus/runtime/graal/NativeSubstitutions.java
tomekl007: We will improve it - see my comments here:
https://datastax-oss.atlassian.net/browse/JAVA-2696
I think that `Substitutions` should be used only if there is no other way to tackle the problem.
For the `isCurrentTimeMicrosAvailable` we have `force_java_clock` setting so it should be ok to use it.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/AbstractCassandraClientProducer.java
tomekl007: `ProgrammaticDriverConfigLoaderBuilder` is a mutable object and has side-effects, that is why this method needs to work that way.

deployment/src/test/java/com/datastax/oss/quarkus/CassandraClientConfigTest.java
tomekl007: The purpose of this test is testing loading data from `resources` folder, so generation would not be a good fit here.

integration-tests/src/main/java/com/datastax/oss/quarkus/dao/nameconverters/NameConverterEntity.java
tomekl007: It may be not a good solution, because in the future we may add more tests that validate other reflection calls - if we will rename it, in such a case there will be name clash. 
I think that the reader should start analyzing this code from:
```
@RegisterForReflection
public class TestNameConverter implements NameConverter 
```
and should notice the `@RegisterForReflection` that will give more insights about the purpose of this class.

cassandra-quarkus-extension/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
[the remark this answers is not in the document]
tomekl007: Yes, I was also thinking about loading this class using IO but I think that it is a bit fragile. 
When I was analyzing other extensions in the `quarkus` project, none of them is loading class this way. They tend to generate it. 

One important note here, when I was debugging it, it seems that quarkus is not creating a physical file with `CassandraClientProducer` class - you will not find this file in any `/target` directory. It seems that the generated content is kept only in memory (and discarded later)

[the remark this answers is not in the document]
tomekl007: I am creating a new ticket for possible improvement of this code in the future:
https://datastax-oss.atlassian.net/browse/JAVA-2722

tomekl007:
All pending comments were addressed as separate JIRA tickets, see:
https://datastax-oss.atlassian.net/browse/JAVA-2681

cassandra-quarkus-extension/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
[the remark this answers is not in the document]
absurdfarce: To be clear: I'm not necessarily saying it's a good idea to solve this problem in this way.  It very well _might_ be but that isn't immediately clear to me.  I _could_ see the pattern of providing a build item based on an existing class file as being something which could possibly be re-used.  And while it's true that the file in question isn't being created at deployment time it shouldn't be an issue since it's already contained in the extension itself.

PR #2, opened 2020-04-07
JAVA-2694: Add Metrics to the extension

tomekl007:
This is the version with a translation between the java-driver dropwizard metrics and microprofile metrics used in the quarkus. 
To see the alternative approach look at this PR: https://github.com/datastax/cassandra-quarkus/pull/3

I think that wrapping has a lot less complexity so I don't know if we should pay the price of adding that complexity to get rid of one dependency.


runtime/src/main/java/com/datastax/oss/quarkus/runtime/metrics/MetricsFinder.java
```diff
@@ -0,0 +1,44 @@
...
+
+    Optional<Metric> metric = driverMetrics.get().getSessionMetric(sessionMetric);
+
+    if (!metric.isPresent()) {
+      throw new IllegalArgumentException(
+          String.format(
+              "Session metric for name: %s is not present in the driver metrics, but should be.",
+              metric));
+    }
+    return metric.get();
```
absurdfarce, line 42:
```suggestion
    return driverMetrics.flatMap(m -> (Optional<Metric>)m.getSessionMetric(sessionMetric)).orElseThrow(() -> {

      return new IllegalArgumentException(
              String.format(
                      "Session metric for name: %s is not present in the driver metrics, but should be.",
                      sessionMetric.getPath()));
    });
```

deployment/src/main/java/com/datastax/oss/quarkus/deployment/MetricsBuildStep.java
```diff
@@ -0,0 +1,161 @@
...
+            .withType(MetricType.COUNTER)
+            .build());
+
+    supportedSessionMetrics.put(
+        CQL_PREPARED_CACHE_SIZE,
+        Metadata.builder()
+            .withName(CQL_PREPARED_CACHE_SIZE.getPath())
+            .withDescription("The size of the driver-side cache of CQL prepared statements.")
+            .withType(MetricType.GAUGE)
+            .build());
```
absurdfarce, line 125: There's _a lot_ of code duplication above.  Can reduce this by iterating over DefaultSessionMeric.values()

deployment/src/main/java/com/datastax/oss/quarkus/deployment/MetricsBuildStep.java
```diff
@@ -0,0 +1,161 @@
...
+    if (typeRaw.equals(MetricType.GAUGE)) {
+      metrics.produce(
+          new MetricBuildItem(metadata, new CassandraGauge(sessionMetric), true, CONFIG_ROOT));
+    } else if (typeRaw.equals(MetricType.METERED)) {
+      metrics.produce(
+          new MetricBuildItem(metadata, new CassandraMetered(sessionMetric), true, CONFIG_ROOT));
+    } else if (typeRaw.equals(MetricType.COUNTER)) {
+      metrics.produce(
+          new MetricBuildItem(metadata, new CassandraCounter(sessionMetric), true, CONFIG_ROOT));
+    }
```
absurdfarce, line 159: Don't need equals() here; you're comparing enum vals here so == is fine.  More directly: you're taking action based on an enum value which seems to cry out for a switch statement instead.

tomekl007:
I think we will choose this solution: https://github.com/datastax/cassandra-quarkus/pull/3#issuecomment-611848970, so I will not be addressing comments in this PR

PR #3, opened 2020-04-07
JAVA-2694: Add Metrics to the extension (reimplement metrics)

tomekl007:
This is the version with a reimplemented driver metrics to use microprofile metrics used in the quarkus. 
To see the alternative approach look at this PR: https://github.com/datastax/cassandra-quarkus/pull/2

I see two cons of this solution:
- I think that the overall complexity of this version is higher than wrapping (look at lines of code, it is 2x and I didn't add tests for implemented metrics yet).
- the timer metrics that we are using in the driver codahale metrics is a custom HdrReservoir that is a `com.codahale.metrics.Reservoir`, so it has a dependency on the codahale. We are not able to use it when excluding dependency. I don’t think that re-implementing HdrReservoir is a good idea.
Even worse than that, I don’t see a way for org.eclipse.microprofile.metrics.MetricRegistry to provide custom Timer implementation (like we do for HdrReservoir) so we would need to strip that functionality and ignore properties as such: 
```
DefaultDriverOption.METRICS_SESSION_CQL_REQUESTS_HIGHEST,
DefaultDriverOption.METRICS_SESSION_CQL_REQUESTS_DIGITS,
DefaultDriverOption.METRICS_SESSION_CQL_REQUESTS_INTERVAL);
```

Only one con is the ability to exclude the `<artifactId>metrics-core</artifactId>`:
https://github.com/datastax/cassandra-quarkus/pull/3/files#diff-48c560bbb8f61d82574c2887b65b498fR68-R71


runtime/src/main/java/com/datastax/oss/quarkus/runtime/metrics/CassandraCounter.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  @Override
+  public void inc(long n) {
+    // no-op: this is read only
+  }
+
+  @Override
+  public long getCount() {
+    Metric metrics = MetricsFinder.getMetrics(sessionMetric);
```
tomekl007, line 52: we still need to wrap, because this needs to be retrieved at the runtime.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/metrics/MicroProfileMetricsUpdater.java
```diff
@@ -0,0 +1,92 @@
...
+    if (isEnabled(metric, profileName)) {
+      // Just initialize eagerly so that the metric appears even when it has no data yet
+      registry.counter(buildFullName(metric, profileName));
+    }
+  }
+
+  protected void initializeTimer(MetricT metric, DriverExecutionProfile config) {
+    String profileName = config.getName();
+    if (isEnabled(metric, profileName)) {
+      String fullName = buildFullName(metric, profileName);
```
tomekl007, line 87: We need to drop `HdrHistogram` support (see my comment in the PR description for more info)

runtime/src/main/java/com/datastax/oss/quarkus/runtime/metrics/CassandraGauge.java
```diff
@@ -0,0 +1,54 @@
...
+  public Long getValue() {
+    Metric metrics = MetricsFinder.getMetrics(sessionMetric);
+    if (!(metrics instanceof Gauge)) {
+      throw new IllegalArgumentException(
+          String.format(
+              "The metric for metric name: %s should be of %s type, but is: %s.",
+              sessionMetric, Gauge.class.getName(), metrics.getClass().getName()));
+    }
+
+    return ((Gauge<Number>) metrics).getValue().longValue();
```
olim7t, line 52: Related to my comment in `MicroProfileMetricsFactory`: the reason you need this custom gauge is because right now the driver's metric reporters do not publish their metrics in the right registry: they use a standalone instance instead of the container's.
If we fix that, this class (and its sister classes for meters and counters) go away.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/metrics/MicroProfileMetricsFactory.java
```diff
@@ -0,0 +1,136 @@
...
+        parseSessionMetricPaths(config.getStringList(DefaultDriverOption.METRICS_SESSION_ENABLED));
+    this.enabledNodeMetrics =
+        parseNodeMetricPaths(config.getStringList(DefaultDriverOption.METRICS_NODE_ENABLED));
+
+    if (enabledSessionMetrics.isEmpty() && enabledNodeMetrics.isEmpty()) {
+      LOG.debug("[{}] All metrics are disabled.", logPrefix);
+      this.registry = null;
+      this.sessionUpdater = NoopSessionMetricUpdater.INSTANCE;
+    } else {
+      this.registry = new MetricsRegistryImpl();
```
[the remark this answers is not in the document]
tomekl007: Yes, I was also thinking about injecting it there, but I thought that this Hdr custom implementation is a blocker. Since you are ok with duplicating `HdrHistogram` I went that path and implemented your suggestions.
Right now the code uses injected `MetricRegistry`: https://github.com/datastax/cassandra-quarkus/pull/3/files#diff-2cda2a7b5c1ea59242c6192fe0b13d95R62
I've also followed the suggestions on the quarkus-extensions docs, and when metrics library is not present, I am injecting `NoopMetricRegistry`:
https://github.com/datastax/cassandra-quarkus/pull/3/files#diff-4d87f287c23fd7e2ebc1d0421519bf60R220

One big advantage of this approach is that node-level metrics now work out-of-the-box.
I've added more tests that cover all metrics (node and session):
https://github.com/datastax/cassandra-quarkus/pull/3/files#diff-2db56229f70156488ab248f0b4822378R56-R93

>SmallRye's DefaultTimer has a public constructor that takes a reservoir. We already know how to write a reservoir backed by HdrHistogram.

Regarding Timer, I think you meant: `TimerImpl` class and `Reservoir`.
I've created a separate JIRA task for this:  
https://datastax-oss.atlassian.net/browse/JAVA-2734

absurdfarce:
I'm kinda conflicted here.  I agree there is more complexity here, although the changes in response to @olim7t 's points have reduced that considerably.  In exchange for that complexity, though, we get some non-trivial benefits.

The first thing that struck me is that the conversion to MicroProfile isn't really tied to Quarkus.  We could (semi-)easily put all this in java-driver and allow any user to use MicroProfile instead of DropWizard for _any_ application.  That may not have been a feature we would've developed in isolation but when we already have it perhaps we should consider it.

A side effect of the above: Quarkus is largely ignorant of this impl: more specifically there aren't any MetricBuildItems getting passed around here.  That in turn means we aren't iterating over all the supported metrics and generating a MetricBuildItem for them.  This seems like a pretty big goodness.  There is some metric-specific stuff in MicroProfile*MetricUpdater but that's not really different from Dropwizard*MetricUpdater in java-driver.

Another observation: we were discussing doing the wrapped impl now and moving to the more robust implementation at some later date.  But if we have the more robust implementation now... is there an argument for waiting?

I am somewhat concerned about effectively maintaining two metrics impls but I'm not sure that's enough to outweigh the benefits described above.

olim7t:
> The first thing that struck me is that the conversion to MicroProfile isn't really tied to Quarkus. We could (semi-)easily put all this in java-driver and allow any user to use MicroProfile instead of DropWizard for any application.

If the driver was still in beta that would definitely be something to investigate. But we're going to hit backward compatibility issues if we do it now, I don't think it's worth the hassle.

At least this PR confirms that the driver's internal design works as intended: you can switch metric frameworks with very little surface changes (namely, `session.getMetrics()` is empty and you have to find another way to expose the metrics).

> A side effect of the above: Quarkus is largely ignorant of this impl: more specifically there aren't any MetricBuildItems getting passed around here.

Yes, I like that too. The only code that needs to be aware of the actual list of driver metrics are the custom `*MetricsUpdater` components.
And as @tomekl007 noted, another huge win is that dynamic node metrics work out of the box.

> I am somewhat concerned about effectively maintaining two metrics impls

I don't think we want to maintain both of them, the goals of the two PRs was to prototype so we can pick one.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/metrics/MicroProfileMetricsFactory.java
```diff
@@ -0,0 +1,122 @@
...
+    } else {
+      this.registry = metricRegistry;
+      this.sessionUpdater =
+          new MicroProfileSessionMetricUpdater(enabledSessionMetrics, registry, context);
+    }
+  }
+
+  @Override
+  public Optional<Metrics> getMetrics() {
+    throw new NotImplementedException("getMetrics() is not supported by the MicroProfile");
```
olim7t, line 76: My first instinct was to make this return `Optional.empty()`, but I think you're right, an exception is more precise.

Is using commons-lang's `NotImplementedException` a common thing in Quarkus? I think `UnsupportedOperationException` would be clear enough.

Also maybe tell users where to look if they want to access the metrics programmatically:
```
getMetrics() is not supported in Quarkus. The driver publishes its metrics directly to the vendor MetricRegistry of the container
```

absurdfarce: The empty option type is returned in cases where metrics are disabled, though... is this case really _that_ different?  I could see a pretty credible argument for a log line containing the info above + a return of an empty option.

If we do decide to go with an exception 👍 to UOE... unless the NotImplemented thing really is common.

absurdfarce:
> If you're talking about dropwizard in the driver core vs. microprofile here, yes there is some duplication and we'll have to remember to update this extension when the driver adds new metrics.

Yeah, this is what I'm concerned about.  I'm not super-enthusiastic about maintaining two distinct metrics implementations let alone two impls in completely different repos.  That was one of my reasons for making the observation about possibly moving this to java-driver; at least there we'd have everything under one roof.

Real as that concern is I think I'm 👍 to @olim7t 's comments earlier... it seems like the approach in this PR offers too many benefits to ignore and should be the approach we take here.

absurdfarce:
Somewhat off-topic, and I don't want to make too big a deal out of it... but we _could_ support programmatic access to metrics with a custom Session type here, right?  It's not at all clear to me that it's worth it, more musing out loud.

tomekl007:
> Somewhat off-topic, and I don't want to make too big a deal out of it... but we _could_ support programmatic access to metrics with a custom Session type here, right? It's not at all clear to me that it's worth it, more musing out loud.

There is programmatic access to metrics via `MetricRegistry`:
```
 @Inject
  @RegistryType(type = MetricRegistry.Type.VENDOR)
  MetricRegistry registry;
```

PR #4, opened 2020-04-09
JAVA-2692: Enable CI on the quarkus extension repository

tomekl007:


Jenkinsfile
```diff
@@ -0,0 +1,184 @@
...
+
+    if [ "${JABBA_VERSION}" != "1.8" ]; then
+      SKIP_JAVADOCS=true
+    else
+      SKIP_JAVADOCS=false
+    fi
+
+    printenv | sort
+    
+    mvn -B -V verify -Dnative -Dmaven.javadoc.skip=${SKIP_JAVADOCS}
```
adutra, line 79: If I remember correctly, the way I created the `native` profile it will execute both non-native and native tests. So here we are probably executing the native tests twice, and building a native image twice.

pom.xml
```diff
@@ -189,6 +194,24 @@ limitations under the License.]]></inlineHeader>
...
+          <execution>
+            <goals>
+              <goal>prepare-agent</goal>
+            </goals>
+          </execution>
+          <execution>
+            <id>report</id>
+            <phase>prepare-package</phase>
+            <goals>
+              <goal>report</goal>
```
adutra, line 210: It would be nice to exclude the mapper generated classes such as `ProductHelper__MapperGenerated` as they are pulling the code coverage down.
You can also exclude test classes like CassandraEndpoint, CassandraNameConverterEndpoint, etc.
And finally, you can also exclude Quarkus generated classes like `CassandraClientBuildTimeConfig$$accessor`.
We can do this in a follow-up ticket.

Jenkinsfile
```diff
@@ -0,0 +1,147 @@
...
+  }
+}
+
+def executeTestsNative() {
+  sh label: 'Execute tests Native', script: '''#!/bin/bash -le
+    # Load CCM environment variables
+    
+    . ${JABBA_SHELL}
+    jabba use ${JABBA_VERSION}
+    jabba use ${GRAALVM_VERSION}
```
emerkle826, line 50: Line 49
```sh
jabba use ${JABBA_VERSION}
```
is unnecessary because line 50 will replace the JDK with the GRAALVM_VERSION

PR #6, opened 2020-04-16
JAVA-2707: Substitute optional dependencies with Quarkus Substitiution

tomekl007:


bom-runtime/pom.xml
```diff
@@ -58,11 +58,7 @@
         <groupId>org.lz4</groupId>
         <artifactId>lz4-java</artifactId>
         <version>1.6.0</version>
-      </dependency>
-      <dependency>
```
adutra, line 62: Why remove snappy completely?

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientBuildTimeConfig.java
```diff
@@ -48,4 +48,17 @@
    */
   @ConfigItem(name = "metrics.node-enabled", defaultValue = "")
   public Optional<List<String>> metricsNodeEnabled;
+
+  /**
+   * The name of the algorithm used to compress protocol frames.
+   *
+   * <ul>
+   *   <li>lz4: requires net.jpountz.lz4:lz4 in the classpath.
+   *   <li>snappy: requires org.xerial.snappy:snappy-java in the classpath. Works only for not
```
adutra, line 57: Nit: "Works only when not in Native mode".

bom-runtime/pom.xml
absurdfarce: I think this came up in standup on Friday... @tomekl007 can correct me if I get something wrong but my recollection is that lz4 is a pure Java impl while the snappy impl relies heavily on the native libraries, which seemed to lead back down the path of JNI issues.  Several other plugins wound up excluding this for similar reasons.

tomekl007: snappy is not supported in quarkus - other extensions are excluding this as well, for example, kafka-extension. Because it is not supported, there is no need to declare snappy dependency at all.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
```diff
@@ -109,6 +110,21 @@
         new ReflectiveClassBuildItem(true, true, ThreadLocalTimestampGenerator.class.getName()));
   }
 
+  @BuildStep
+  List<ReflectiveClassBuildItem> registerLz4ForReflection(
+      CassandraClientBuildTimeConfig buildTimeConfig) {
+    if (buildTimeConfig.protocolCompression.equalsIgnoreCase("lz4")) {
+      return Arrays.asList(
+          new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4Compressor"),
```
adutra, line 118: Out of curiosity how to do establish this list? Is it by trial and error? How can we be sure that we are not forgetting a class here?

tomekl007: I did it by trial and error. `Lz4` is working now (IT test is covering it) so I think there is no way that we forgot any class, right?

absurdfarce: I guess I'm more concerned that we might be including/referencing some things we don't need to here.

Tests still pass even if I remove one of those dependencies:

```
diff --git a/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java b/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
index 70be4e2..16c920f 100644
--- a/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
+++ b/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
@@ -118,8 +118,8 @@ class CassandraClientProcessor {
           new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4Compressor"),
           new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4JavaSafeCompressor"),
           new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4HCJavaSafeCompressor"),
-          new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4JavaSafeFastDecompressor"),
-          new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4JavaSafeSafeDecompressor"));
+          new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4JavaSafeFastDecompressor"));
+      // new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4JavaSafeSafeDecompressor"));
     } else {
       return Collections.emptyList();
     }
```

Is it worth taking a look at the lz4 impl and seeing if we can deduce what classes will be needed?  Or are classes being included/used based on various configuration/runtime options which aren't being exercised here?

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientBuildTimeConfig.java
```diff
@@ -48,4 +48,17 @@
...
+   *
+   * <ul>
+   *   <li>lz4: requires net.jpountz.lz4:lz4 in the classpath.
+   *   <li>snappy: requires org.xerial.snappy:snappy-java in the classpath. Works only when not in
+   *       Native mode.
+   *   <li>the string "none" to indicate no compression - this is a default value.
+   * </ul>
+   */
+  @ConfigItem(name = "protocol.compression", defaultValue = "none")
+  public String protocolCompression;
```
absurdfarce, line 63: Do we want this to _only_ be specified at build time?  If we also make this support a run-time override we could add this to the runtime props in CassandraClientConnectionConfig (which seems like a fairly natural home for it anyways) and just set the value on the runtime config we pass as an arg to CassandraClientProcessor.configureRuntimeProperties().  The main advantage I see is that this would allow us to avoid the per-operation logic (recorder.configureCompression, recorder.configureMetricsRegistry, recoder.configureRuntimeProps); it seems like the way we're doing it now could snowball in complexity as we add new things to support here.

In short, I'm wondering if we can't consolidate most of this in a single representation (which naturally would be the runtime config) and avoid a fair amount of the ad-hoc CassandraClientRecorder customization here.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
```diff
@@ -205,7 +229,14 @@ void configureRuntimeProperties(
...
       Capabilities capabilities) {
     recorder.configureRuntimeProperties(runtimeConfig);
+    configureMetrics(recorder, buildTimeConfig, capabilities);
+    recorder.configureCompression(buildTimeConfig.protocolCompression);
+  }
 
+  private void configureMetrics(
+      CassandraClientRecorder recorder,
+      CassandraClientBuildTimeConfig buildTimeConfig,
+      Capabilities capabilities) {
```
absurdfarce, line 239: Maybe it's just me but I find the relationship between the processor and the recorder to be somewhat unclear.  I'm wondering if we shouldn't more clearly define roles here, specifically:

* The recorder exposes methods which take raw bits of data (most likely config data) and constructs what it needs internally to subsequently necessary to create CqlSessions
* The processor evaluates conditions at runtime and extracts/computes/otherwise obtains those raw bits of data.  It then uses the methods supplied by the recorder to setup what is needed.

Specifically, the processor shouldn't be responsible for constructing things like MetricsConfig instances.  It's job is to tell the recorder to either enable metrics (for a specified set of node and/or session metrics) or disable metrics.  The recorder is responsible for actually implementing that logic.  I'm envisioning something more like the following:

```java
  @Record(RUNTIME_INIT)
  @BuildStep
  void configureRuntimeProperties(
      CassandraClientRecorder recorder,
      CassandraClientConfig runtimeConfig,
      CassandraClientBuildTimeConfig buildTimeConfig,
      Capabilities capabilities) {

    recorder.runtimeProperties(runtimeConfig).setCompression(buildTimeConfig.protocolCompression);

    if (buildTimeConfig.metricsEnabled) {
      recorder.enableMetrics(
          buildTimeConfig.metricsNodeEnabled, buildTimeConfig.metricsSessionEnabled);
    } else {
      recorder.disableMetrics();
    }

    if (buildTimeConfig.metricsEnabled && capabilities.isCapabilityPresent(Capabilities.METRICS)) {
      recorder.setInjectedMetricRegistry();
    } else {
      recorder.setNoopMetricRegistry();
    }
  }
```

I've added some method chaining to further simplify the code but you get the general idea I'm going for here.

My suspicion is that as we add support for additional features a clear delineation of roles here will help us keep track of what's going on in this code.  Note that this is separate from the question of whether protocol compression should even have it's own method in the recorder vs. just being added to CassandraClientConnectionConfig.

Full diff supporting the snippet I showed above follows:

```
diff --git a/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java b/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
index 70be4e2..3ac96cb 100644
--- a/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
+++ b/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
@@ -65,7 +65,6 @@ import io.quarkus.smallrye.health.deployment.spi.HealthBuildItem;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
-import java.util.Optional;
 import javax.enterprise.context.ApplicationScoped;
 import javax.enterprise.inject.Default;
 import javax.enterprise.inject.Produces;
@@ -228,21 +227,14 @@ class CassandraClientProcessor {
       CassandraClientConfig runtimeConfig,
       CassandraClientBuildTimeConfig buildTimeConfig,
       Capabilities capabilities) {
-    recorder.configureRuntimeProperties(runtimeConfig);
-    configureMetrics(recorder, buildTimeConfig, capabilities);
-    recorder.configureCompression(buildTimeConfig.protocolCompression);
-  }
 
-  private void configureMetrics(
-      CassandraClientRecorder recorder,
-      CassandraClientBuildTimeConfig buildTimeConfig,
-      Capabilities capabilities) {
+    recorder.runtimeProperties(runtimeConfig).setCompression(buildTimeConfig.protocolCompression);
+
     if (buildTimeConfig.metricsEnabled) {
-      recorder.configureMetrics(
-          new MetricsConfig(
-              buildTimeConfig.metricsNodeEnabled, buildTimeConfig.metricsSessionEnabled, true));
+      recorder.enableMetrics(
+          buildTimeConfig.metricsNodeEnabled, buildTimeConfig.metricsSessionEnabled);
     } else {
-      recorder.configureMetrics(new MetricsConfig(Optional.empty(), Optional.empty(), false));
+      recorder.disableMetrics();
     }
 
     if (buildTimeConfig.metricsEnabled && capabilities.isCapabilityPresent(Capabilities.METRICS)) {
diff --git a/runtime/src/main/java/com/datastax/oss/quarkus/runtime/CassandraClientRecorder.java b/runtime/src/main/java/com/datastax/oss/quarkus/runtime/CassandraClientRecorder.java
index 1efbf1c..53f8d0e 100644
--- a/runtime/src/main/java/com/datastax/oss/quarkus/runtime/CassandraClientRecorder.java
+++ b/runtime/src/main/java/com/datastax/oss/quarkus/runtime/CassandraClientRecorder.java
@@ -24,6 +24,8 @@ import io.quarkus.arc.runtime.BeanContainerListener;
 import io.quarkus.runtime.RuntimeValue;
 import io.quarkus.runtime.annotations.Recorder;
 import io.smallrye.metrics.MetricRegistries;
+import java.util.List;
+import java.util.Optional;
 import javax.enterprise.inject.Default;
 import javax.enterprise.util.AnnotationLiteral;
 import org.eclipse.microprofile.metrics.MetricRegistry;
@@ -36,10 +38,11 @@ public class CassandraClientRecorder {
     return beanContainer -> beanContainer.instance(cqlSessionProducerClass);
   }
 
-  public void configureRuntimeProperties(CassandraClientConfig config) {
+  public CassandraClientRecorder runtimeProperties(CassandraClientConfig config) {
     AbstractCassandraClientProducer producer =
         Arc.container().instance(AbstractCassandraClientProducer.class).get();
     producer.setCassandraClientConfig(config);
+    return this;
   }
 
   private AnnotationLiteral<Default> defaultName() {
@@ -50,10 +53,23 @@ public class CassandraClientRecorder {
     return new RuntimeValue<>(Arc.container().instance(CqlSession.class, defaultName()).get());
   }
 
-  public void configureMetrics(MetricsConfig metricsConfig) {
-    AbstractCassandraClientProducer producer =
+  public CassandraClientRecorder enableMetrics(
+      Optional<List<String>> nodeMetrics, Optional<List<String>> sessionMetrics) {
+
+    return setMetricConfig(new MetricsConfig(nodeMetrics, sessionMetrics, true));
+  }
+
+  public CassandraClientRecorder disableMetrics() {
+
+    return setMetricConfig(new MetricsConfig(Optional.empty(), Optional.empty(), false));
+  }
+
+  private CassandraClientRecorder setMetricConfig(MetricsConfig config) {
+
+    AbstractCassandraClientProducer rv =
         Arc.container().instance(AbstractCassandraClientProducer.class).get();
-    producer.setMetricsConfig(metricsConfig);
+    rv.setMetricsConfig(config);
+    return this;
   }
 
   public void setInjectedMetricRegistry() {
@@ -69,7 +85,7 @@ public class CassandraClientRecorder {
     producer.setMetricRegistry(new NoopMetricRegistry());
   }
 
-  public void configureCompression(String protocolCompression) {
+  public void setCompression(String protocolCompression) {
     AbstractCassandraClientProducer producer =
         Arc.container().instance(AbstractCassandraClientProducer.class).get();
     producer.setProtocolCompression(protocolCompression);
```

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
absurdfarce: Unfortunately the only way to really confirm that @tomekl007 is to look at the loading code.  It's entirely possible (arguably even likely) that the logic may take different paths based on env vars, sys props or any other test and that that logic might play differently on other systems.  That was certainly true for jnr and it looks like it's true here.... although I think what you have now will work as the base (if not optimized) case.

The driver asks the lz4 impl for the fastest instance it can come up with (https://github.com/datastax/java-driver/blob/2863100b736f01b77f7347d17536da6f151b4583/core/src/main/java/com/datastax/oss/driver/internal/core/protocol/Lz4Compressor.java#L39).  This will actually try to load a native (JNI) instance, falling back onto a pure Java instance if that fails (which it will in our case) (https://github.com/lz4/lz4-java/blob/1.6.0/src/java/net/jpountz/lz4/LZ4Factory.java#L160-L171).

There are two Java instances to choose from, one using Unsafe and the other a "safe" instance.  The loader code will try the Unsafe version first (assuming it looks like Unsafe is, er, "safe" on the current architecture) and fall back to the "safe" version if that fails (https://github.com/lz4/lz4-java/blob/1.6.0/src/java/net/jpountz/lz4/LZ4Factory.java#L136-L146).  The actual test to see if Unsafe is supported is pretty straightforward (https://github.com/lz4/lz4-java/blob/1.6.0/src/java/net/jpountz/util/Utils.java#L25-L34); it returns true for my FC28 instance and I'm guessing it returns true for you as well.  Because the "JavaUnsafe" classes _aren't_ included via the reflection config, however, you get a CNFE exception when trying to load them... which will cause the loader code to return the "JavaSafe" versions as the baseline default.  That's why you saw those classes referenced explicitly in your testing and that's why including them makes the whole thing work.

To validate this, I confirmed that native integration tests pass with the following diff applied:

```
diff --git a/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java b/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
index 70be4e2..965f901 100644
--- a/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
+++ b/deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
@@ -116,10 +116,11 @@ class CassandraClientProcessor {
     if (buildTimeConfig.protocolCompression.equalsIgnoreCase("lz4")) {
       return Arrays.asList(
           new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4Compressor"),
-          new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4JavaSafeCompressor"),
-          new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4HCJavaSafeCompressor"),
-          new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4JavaSafeFastDecompressor"),
-          new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4JavaSafeSafeDecompressor"));
+          new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4JavaUnsafeCompressor"),
+          new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4HCJavaUnsafeCompressor"),
+          new ReflectiveClassBuildItem(true, true, "net.jpountz.lz4.LZ4JavaUnsafeFastDecompressor"),
+          new ReflectiveClassBuildItem(
+              true, true, "net.jpountz.lz4.LZ4JavaUnsafeSafeDecompressor"));
     } else {
       return Collections.emptyList();
     }
```

All of this leads me to one comment and one question:

* We should probably include both the "JavaSafe" and "JavaUnsafe" classes in the reflection config.  That way if a user has the ability to leverage the (faster) Unsafe impl they can do so
* The initial call from driver code will try to load a JNI impl... but we really don't want to do that in a Graal context.  Should we consider changing this behaviour (perhaps via a broader substitution class below) to just call Lz4Factory.fastestJavaInstance() in the Graal case?

tomekl007: Thanks for your investigation! 
- I agree that we should provide both versions, done in https://github.com/datastax/cassandra-quarkus/pull/6/commits/192657b3c7cd548d8924a673f422f9e5985880f9
- I think that we don't need to alter (substitute) this logic. It seems that the Lz4 Loader logic is written in that way to work exactly in such cases. If it works, I think we should not add additional complexity to remove only one method call, wdyt?

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientBuildTimeConfig.java
tomekl007: all settings that are in build time are here because this information is needed at compile time. Also, according to the quarkus team, we should move as many settings as possible to build time (instead of runtime). Regarding:
>The main advantage I see is that this would allow us to avoid the per-operation logic (recorder.configureCompression, recorder.configureMetricsRegistry, recoder.configureRuntimeProps);

If that will become an issue, we can refactor setting build time config into common code. Right now we have only two methods so I think it will be over-design at this point of the project.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
tomekl007: Actually, this is another way around. The processor logic must be in the build time.
I've taken then the pattern of Recorder, Processor and their responsibilities from other extensions. Regarding returns `this`, I don't see a clear benefit of doing this. The recorder is not a builder. It is a way to integrate the build time code with runtime.

absurdfarce: I'm not sure I understand what you mean @tomekl007.  Nothing in the build-time/run-time distinction forces you to build MetricsConfig instances in the processor rather than the recorder.  In fact you're already doing what I'm suggesting with the metric registry.  In that case you call a method on the recorder and that method is responsible for actually creating the instance.  All I'm suggesting is making this pattern uniform.

Returning this is primarily about simplifying the code.  It's something I'd argue for personally but it's not essential by any means.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientBuildTimeConfig.java
absurdfarce: I don't think we _need_ the information at build-time.  We don't do anything with it until we use the build-time config info to populate the PDCLB... but that happens at run-time.  It seems like we're moving it to build-time because we want to follow the Quarkus suggestion to support as much as we can at build-time... and I agree with that.

I'm not arguing to move it to run-time only.   I'm suggesting we could specify it at build-time and support a run-time override.  This approach would allow us to support something even more powerful; a unified configuration that lives in a single place.  That's really what I'm after.

This isn't necessarily something we need to address in this PR; I know there's other work to cover other bits of the configuration.  But it is something I'd argue we need to consider at some point in the future.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
absurdfarce: I'm fine with the change to include the unsafe variants.  And I mostly agree with you about not bothering with the shift to Lz4Factory.fastestJavaInstance() for Quarkus, at least for now.  The users that would actually pay the penalty of trying the native code first is (a) Graal users who are (b) also using compression... which can't be a huge population of users.  And those users pay a one-time cost that isn't that big anyway.  I'm not convinced it's worth the engineering effort to fix, and even if we do it should be fixed in the Java driver (using an if block based on whether or no we're running in a Graal VM)... it's not something we should fix with a substitution.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientBuildTimeConfig.java
tomekl007: We need that info on build time for registering reflection classes, see:
https://github.com/datastax/cassandra-quarkus/pull/6/files#diff-4d87f287c23fd7e2ebc1d0421519bf60R116

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
tomekl007: I think that we can continue the discussion about settings in this ticket:
https://datastax-oss.atlassian.net/browse/JAVA-2697
Please add your thought about shape of the configuration mechanism there.

PR #7, opened 2020-04-23
JAVA-2736: Speed up TestContainers startup time

emerkle826:
There are a few configurations you can tweak to make Cassandra startup faster (disabling gossip and vnodes, for example). With TestContainers, the ability to control this is a little limiting. It allows you to provide a configuration override when creating the container, but the override is simply a directory that is mapped to the Cassandra config directory (/etc/cassandra inside the container).

With this approach, settings can be tweaked, but you have to provide **entire** files for anything you wish to override: you can't simply provide partial overrides. This PR incorporates a few tweaks borrowed from  [here](https://github.com/sky-uk/cassandra-operator/issues/103) that result in about a 25 - 30% reduction in total Integration test time (locally at least).

### DIFF
--- a/deployment/src/test/resources/test-config/cassandra-env.sh
+++ b/deployment/src/test/resources/test-config/cassandra-env.sh
@@ -1,0 +1,312 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+calculate_heap_sizes()
+{
+    case "`uname`" in
+        Linux)
+            system_memory_in_mb=`free -m | awk '/:/ {print $2;exit}'`
+            system_cpu_cores=`egrep -c 'processor([[:space:]]+):.*' /proc/cpuinfo`
+        ;;
+        FreeBSD)
+            system_memory_in_bytes=`sysctl hw.physmem | awk '{print $2}'`
+            system_memory_in_mb=`expr $system_memory_in_bytes / 1024 / 1024`
+            system_cpu_cores=`sysctl hw.ncpu | awk '{print $2}'`
+        ;;
+        SunOS)
+            system_memory_in_mb=`prtconf | awk '/Memory size:/ {print $3}'`
+            system_cpu_cores=`psrinfo | wc -l`
+        ;;
+        Darwin)
+            system_memory_in_bytes=`sysctl hw.memsize | awk '{print $2}'`
+            system_memory_in_mb=`expr $system_memory_in_bytes / 1024 / 1024`
+            system_cpu_cores=`sysctl hw.ncpu | awk '{print $2}'`
+        ;;
+        *)
+            # assume reasonable defaults for e.g. a modern desktop or
+            # cheap server
+            system_memory_in_mb="2048"
+            system_cpu_cores="2"
+        ;;
+    esac
+
+    # some systems like the raspberry pi don't report cores, use at least 1
+    if [ "$system_cpu_cores" -lt "1" ]
+    then
+        system_cpu_cores="1"
+    fi
+
+    # set max heap size based on the following
+    # max(min(1/2 ram, 1024MB), min(1/4 ram, 8GB))
+    # calculate 1/2 ram and cap to 1024MB
+    # calculate 1/4 ram and cap to 8192MB
+    # pick the max
+    half_system_memory_in_mb=`expr $system_memory_in_mb / 2`
+    quarter_system_memory_in_mb=`expr $half_system_memory_in_mb / 2`
+    if [ "$half_system_memory_in_mb" -gt "1024" ]
+    then
+        half_system_memory_in_mb="1024"
+    fi
+    if [ "$quarter_system_memory_in_mb" -gt "8192" ]
+    then
+        quarter_system_memory_in_mb="8192"
+    fi
+    if [ "$half_system_memory_in_mb" -gt "$quarter_system_memory_in_mb" ]
+    then
+        max_heap_size_in_mb="$half_system_memory_in_mb"
+    else
+        max_heap_size_in_mb="$quarter_system_memory_in_mb"
+    fi
+    MAX_HEAP_SIZE="${max_heap_size_in_mb}M"
+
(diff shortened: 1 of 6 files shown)


emerkle826:
**NOTE:** The `cassandra.yaml` and `cassandra-env.sh` files I've added in the PR were obtained by spinning up an instance of the Docker container the tests currently use and copying them out.
Also, the default Docker container for TestContainers Cassandra is a 3.11.2 build. If the tests decide to change the Cassandra version, or introduce multiple versions, the approach in this PR may have to change as well.

deployment/src/test/resources/test-config/cassandra-env.sh
```diff
@@ -0,0 +1,312 @@
...
+JVM_OPTS="$JVM_OPTS -Djava.library.path=$CASSANDRA_HOME/lib/sigar-bin"
+
+JVM_OPTS="$JVM_OPTS $MX4J_ADDRESS"
+JVM_OPTS="$JVM_OPTS $MX4J_PORT"
+JVM_OPTS="$JVM_OPTS $JVM_EXTRA_OPTS"
+# TEST CONTAINER SPEED UP
+JVM_OPTS="$JVM_OPTS -Dcassandra.initial_token=0"
+JVM_OPTS="$JVM_OPTS -Dcassandra.skip_wait_for_gossip_to_settle=0"
+JVM_OPTS="$JVM_OPTS -Dcassandra.load_ring_state=false"
+JVM_OPTS="$JVM_OPTS -Dcassandra.allocate_tokens_for_local_replication_factor=nil"
```
emerkle826, line 312: These are the only changes from the `cassandra-env.sh` file that TestContainers provides out of the box (for version 3.11.2 at least)

deployment/src/test/resources/test-config/cassandra.yaml
```diff
@@ -0,0 +1,1241 @@
...
+# If you leave this unspecified, Cassandra will use the default of 1 token for legacy compatibility,
+# and will use the initial_token as described below.
+#
+# Specifying initial_token will override this setting on the node's initial start,
+# on subsequent starts, this setting will apply even if initial token is set.
+#
+# If you already have a cluster with 1 token per node, and wish to migrate to 
+# multiple tokens per node, see http://wiki.apache.org/cassandra/Operations
+# TEST CONTAINER SPEED UP
+#num_tokens: 256
```
emerkle826, line 26: `num_tokens` needs to be commented out to disable vnodes (normally uncommented and set to 256)

deployment/src/test/resources/test-config/cassandra.yaml
```diff
@@ -0,0 +1,1241 @@
...
+#
+# Only supported with the Murmur3Partitioner.
+# allocate_tokens_for_keyspace: KEYSPACE
+
+# initial_token allows you to specify tokens manually.  While you can use it with
+# vnodes (num_tokens > 1, above) -- in which case you should provide a 
+# comma-separated list -- it's primarily used when adding nodes to legacy clusters 
+# that do not have vnodes enabled.
+# TEST CONTAINER SPEED UP
+initial_token: 0
```
emerkle826, line 44: when you disable vnodes, you need to specify the initial token here. (normally commented out)

deployment/src/test/resources/test-config/cassandra.yaml
```diff
@@ -0,0 +1,1241 @@
...
+    # Addresses of hosts that are deemed contact points. 
+    # Cassandra nodes use this list of hosts to find each other and learn
+    # the topology of the ring.  You must change this if you are running
+    # multiple nodes!
+    - class_name: org.apache.cassandra.locator.SimpleSeedProvider
+      parameters:
+          # seeds is actually a comma-delimited list of addresses.
+          # Ex: "<ip1>,<ip2>,<ip3>"
+          # TEST CONTAINER SPEED UP
+          - seeds: "127.0.0.1"
```
emerkle826, line 428: setting the seed to the node itself is recommended for speedup as well (see [here](https://github.com/saidbouras/cassandra-docker-unit/blob/master/scripts/setup-config.sh#L13-L14))

deployment/src/test/resources/test-config/cassandra.yaml
```diff
@@ -0,0 +1,1241 @@
...
+# snapshots for you.  Mostly useful if you're paranoid when there
+# is a data format change.
+snapshot_before_compaction: false
+
+# Whether or not a snapshot is taken of the data before keyspace truncation
+# or dropping of column families. The STRONGLY advised default of true 
+# should be used to provide data safety. If you set this flag to false, you will
+# lose data on truncation or drop.
+# TEST CONTAINER SPEED UP
+auto_snapshot: false
```
emerkle826, line 778: disable snapshots to try to speed things up as well.

integration-tests/src/test/resources/test-config/cassandra-env.sh
```diff
@@ -0,0 +1,312 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
emerkle826, line 1: This is a duplicate of the same file above. Same changes made here.

integration-tests/src/test/resources/test-config/cassandra.yaml
```diff
@@ -0,0 +1,1241 @@
+# Cassandra storage config YAML
```
emerkle826, line 1: Again, this is a duplicate of the same file above, same changes made.

tomekl007:
Wow, this is really a lot of code - I am a bit afraid about further maintenance of it. 
Isn't there a way to configure only differences using programmatic API? (maybe using other `with*` methods)

emerkle826:
> Wow, this is really a lot of code - I am a bit afraid about further maintenance of it.
> Isn't there a way to configure only differences using programmatic API? (maybe using other `with*` methods)

I've looked at the API provided via Testcontainers and it doesn't currently allow for updating parts of file resources. But you have valid points as maintaining these config files is not going to be easy and makes this test approach brittle.

I'll open a new PR to investigate ways to possibly enhance Testcontainers itself to allow for partial config overrides. For now, I'm converting this PR into a Draft PR.

olim7t:
> But you have valid points as maintaining these config files is not going to be easy and makes this test approach brittle.

💯 I too am worried about long-term maintainability, especially as the image gets updated and/or multiple Cassandra versions are supported.

Would the changes be generally beneficial to anyone using TestContainers? We could try to get them into the base images directly, or maybe there is a way to host our own images.

adutra:
Created [JAVA-2781](https://datastax-oss.atlassian.net/browse/JAVA-2781) to pursue with the idea of contributing to TestContainers. I think we close this for now.

adutra:
Also create an umbrella epic link for TestContainers contributions: https://datastax-oss.atlassian.net/browse/JAVA-2787.

deployment/src/test/resources/test-config/cassandra-env.sh
```diff
@@ -0,0 +1,312 @@
...
```
alexott, line 312: It's easier & preferable to put this stuff into `jvm.options`

adutra:
@emerkle826 I think we can close this one for now? We need to improve things on the TestContainers side first, but once that's done, we likely won't need the whole cassandra.yaml file here anyways.

emerkle826:
> @emerkle826 I think we can close this one for now? We need to improve things on the TestContainers side first, but once that's done, we likely won't need the whole cassandra.yaml file here anyways.

Agreed. I'll revisit this after the TestContainers improvements.

PR #8, opened 2020-04-27
JAVA-2683: Create a Quarkus Cassandra quick starter

tomekl007:
todo:
- [postponed] async example - this will be done as a part of https://datastax-oss.atlassian.net/browse/JAVA-2754

### DIFF
--- a/quickstart/README.md
+++ b/quickstart/README.md
@@ -1,0 +1,54 @@
+# Quarkus demo: Cassandra Client
+
+This example showcases how to use the Cassandra client with Quarkus. 
+
+# Cassandra instance - Running with Docker
+
+Just run it as follows:
+`docker run \
+   --name local-cassandra-instance \
+   -p 7000:7000 \
+   -p 7001:7001 \
+   -p 7199:7199 \
+   -p 9042:9042 \
+   -p 9160:9160 \
+   -p 9404:9404 \
+   -d \
+   launcher.gcr.io/google/cassandra3`
+
+**Provision table**
+
+`docker exec -it local-cassandra-instance cqlsh -e "CREATE KEYSPACE IF NOT EXISTS k1 WITH replication = {'class':'SimpleStrategy', 'replication_factor':1}"`
+`docker exec -it local-cassandra-instance cqlsh -e "CREATE TABLE IF NOT EXISTS k1.fruit(id text, name text, description text, PRIMARY KEY((id), name))"`
+                                                     
+
+# Run the demo on dev mode
+
+- Run `mvn clean package` and then `java -jar ./target/quickstart-1.0.0-SNAPSHOT-runner.jar`
+- In dev mode `mvn clean quarkus:dev`
+
+Go to `http://localhost:8080/fruits.html`, it should show a simple App to manage list of Fruits. 
+You can add fruits to the list via the form.
+
+Alternatively, you can use curl commands to interact with the underlying REST API.
+To create a fruit:
+```shell script
+curl --header "Content-Type: application/json" \
+  --request POST \
+  --data '{"name":"curl_product","description":"this was created via curl"}' \
+  http://localhost:8080/fruits
+```
+To retrieve fruits:
+```shell script
+curl -X GET http://localhost:8080/fruits
+```
+
+# Running in native
+
+You can compile the application into a native binary using:
+
+`mvn clean package -Pnative`
+
+and run with:
+
+`./target/quickstart-1.0.0-SNAPSHOT-runner` 
--- a/quickstart/README.md
+++ b/quickstart/README.md
@@ -4,7 +4,7 @@ This example showcases how to use the Cassandra client with Quarkus.
 
 # Cassandra instance - Running with Docker
 
-Just run it as follows:
+Just run it as follows (only 9042 port is required for quickstart application):
 `docker run \
    --name local-cassandra-instance \
    -p 7000:7000 \
--- a/quickstart/src/main/docker/Dockerfile.jvm
+++ b/quickstart/src/main/docker/Dockerfile.jvm
@@ -1,0 +1,22 @@
+####
+# This Dockerfile is used in order to build a container that runs the Quarkus application in JVM mode
+#
+# Before building the docker image run:
+#
+# mvn package
+#
+# Then, build the image with:
+#
+# docker build -f src/main/docker/Dockerfile.jvm -t quarkus/mongo-client-jvm .
+#
+# Then run the container using:
+#
+# docker run -i --rm -p 8080:8080 quarkus/mongo-client-jvm
+#
+###
+FROM fabric8/java-alpine-openjdk8-jre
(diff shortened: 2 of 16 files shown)


quickstart/src/main/docker/Dockerfile.jvm
```diff
@@ -0,0 +1,22 @@
+####
```
absurdfarce, line 1: It's a little odd that we just have these Dockerfiles hanging out here with no obvious context; we talk about running the app locally in the README but there's no mention of running in a container.  Maybe we should at least mention these there?

quickstart/README.md
```diff
@@ -0,0 +1,54 @@
...
+`docker run \
+   --name local-cassandra-instance \
+   -p 7000:7000 \
+   -p 7001:7001 \
+   -p 7199:7199 \
+   -p 9042:9042 \
+   -p 9160:9160 \
+   -p 9404:9404 \
+   -d \
+   launcher.gcr.io/google/cassandra3`
```
absurdfarce, line 17: We don't _need_ to expose all those ports to vet out the quickstart, right?  If we get 9042 we should be okay.

Maybe just add a comment here saying only 9042 is required and the others are optional?

absurdfarce, line 17: I suppose another way to cut down on these ports would be to expose a k8s config that would run C* + quickstart container... that would only need to expose port 8080.  But that's probably beyond the scope of a quickstart app.

quickstart/README.md
```diff
@@ -0,0 +1,54 @@
...
+To retrieve fruits:
+```shell script
+curl -X GET http://localhost:8080/fruits
+```
+
+# Running in native
+
+You can compile the application into a native binary using:
+
+`mvn clean package -Pnative`
```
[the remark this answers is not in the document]
tomekl007: This was fixed in the https://datastax-oss.atlassian.net/browse/JAVA-2707.
Please `mvn install` the `cassandra-quarkus` on your local and start again.

quickstart/README.md
```diff
@@ -0,0 +1,54 @@
...
+   -p 9042:9042 \
+   -p 9160:9160 \
+   -p 9404:9404 \
+   -d \
+   launcher.gcr.io/google/cassandra3`
+
+**Provision table**
+
+`docker exec -it local-cassandra-instance cqlsh -e "CREATE KEYSPACE IF NOT EXISTS k1 WITH replication = {'class':'SimpleStrategy', 'replication_factor':1}"`
+`docker exec -it local-cassandra-instance cqlsh -e "CREATE TABLE IF NOT EXISTS k1.fruit(id text, name text, description text, PRIMARY KEY((id), name))"`
```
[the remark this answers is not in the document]
tomekl007: yes, I wanted to make it work this way, but -f file for docker is not working for me.

quickstart/src/main/resources/META-INF/resources/fruits.html
```diff
@@ -0,0 +1,98 @@
+<!doctype html>
```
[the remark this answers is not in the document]
tomekl007: yes, this is a common UI for other quick starter's as well, i.e: hazelcast, dynamodb

quickstart/src/main/resources/META-INF/resources/index.html
```diff
@@ -0,0 +1,152 @@
...
+        <p>Just delete the <code>src/main/resources/META-INF/resources/index.html</code> file.</p>
+    </div>
+    <div class="right-column">
+        <div class="right-section">
+            <h3>Application</h3>
+            <ul>
+                <li>GroupId: com.datastax.oss.quarkus</li>
+                <li>ArtifactId: cassandra-quarkus-client</li>
+                <li>Version: 1.0-SNAPSHOT</li>
+                <li>Quarkus Version: 1.3.1.Final</li>
```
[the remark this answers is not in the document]
tomekl007: We can use `1.3.1` I've changed it

quickstart/src/main/docker/Dockerfile.jvm
tomekl007: The readme is in the dockerfiles:
https://github.com/datastax/cassandra-quarkus/pull/8/files/51bdfadffad1fda23ab550529de769cc4f20af10#diff-5b11c784af46052926eaa14bfec26b15R2-R16

quickstart/README.md
tomekl007: I think that not all people using starters will be familiar with kubernetes. 
The current setup is easier because it starts the Cassandra on local using docker, and users can interact with it because all ports are exposed.

tomekl007: I wanted to give the users full exposed Cassandra so I can connect using CQLSH, JMX, driver and so one.

quickstart/src/main/docker/Dockerfile.jvm
absurdfarce: Yeah, I saw that @tomekl007 .... I just meant that it seemed odd that the main README talks about running all the apps locally (with no mention of Docker) and then there are separate Dockerfiles provided to do the same thing.  It's not a huge thing it just seemed a little jarring.

quickstart/src/main/resources/META-INF/resources/fruits.html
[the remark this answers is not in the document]
absurdfarce: Ah, got it, okay, thanks for the clarification!

quickstart/README.md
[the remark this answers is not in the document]
absurdfarce: Oh, I see, this is the stuff your substitutions address... got it.  Thanks for the pointer, and I can confirm that a rebuild + install of cassandra-quarkus did fix the issue.

quickstart/README.md
absurdfarce: Yeah, I think that's right @tomekl007 ... for a quickstart approach this seems more straightforward.

quickstart/README.md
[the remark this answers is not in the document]
absurdfarce: "docker exec" will run the command in the container so the file would need to be available.  I think you're looking for something like this:

```
docker run --name local-cassandra-instance -p 9042:9042 -d -v $PWD/quickstart:/quickstart launcher.gcr.io/google/cassandra3
```

Plus or minus the -v syntax here rather than the newer --mount... but you get the point.  $PWD here is the root of the source checkout.  This should get you to a point where you can do:

```
docker exec -it local-cassandra-instance cqlsh -f /quickstart/setup_db.cql
```

I keep running aground trying that here but I'm trying it on FC28 so selinux is stomping on everything I try to do.  Anyway, this might not be worth the complexity.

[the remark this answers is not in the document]
tomekl007: Yes, mounting the volume only for that purpose is too much.

quickstart/src/main/docker/Dockerfile.jvm
tomekl007: I am following the same pattern that is for other quick-starts. They all have those dockerfiles.

PR #9, opened 2020-05-05
JAVA-2684: Add documentation and examples of Quarkus Cassandra extension

tomekl007:
todo:
- [x] add generated config settings to the documentation - blocked, see https://github.com/quarkusio/quarkus/issues/7876#issuecomment-624115140
https://github.com/datastax/cassandra-quarkus/tree/java2684_settings_generation
- [ ] hosting docs, and change `guide: "https://quarkus.io/guides/cassandra"` to existing link


absurdfarce:
Note: asciidoc used here as it's the primary documentation method for Quarkus... see https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/README.adoc

absurdfarce:
It's interesting that most of the Quarkus extensions (or at least those defined at https://github.com/quarkusio/quarkus/tree/master/extensions) don't have per-extension documentation.  It _looks like_ their docs are just managed as part of the main project doc at https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc.  Is this still an ongoing thing?  Should we be making an effort to get our documentation into src/main/asciidoc rather than within the extension itself?

absurdfarce:
I don't love the idea that we're pulling in theme materials from the Quarkus docs here; it seems far too easy for our copy of those materials to get out of sync should the project decide to change that around.  Is there any way we can avoid pulling all this material in... maybe some way of automatically retrieving style/theme materials for doc runs if they aren't already present locally?

absurdfarce:
The JIRA ticket mentioned a CONTRIBUTING doc for this repo as well... is that still something we plan on adding?

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,301 @@
...
+The `pom.xml` is importing the RESTEasy/JAX-RS, JSON-B, Context Propagation and Cassandra Client extensions.
+We will be building a REST application that is using DataStax Object Mapper. It will simplify the Data Access Layer code.
+
+The most important part of the `pom.xml` is adding the `cassandra-quarkus` extension:
+[source,xml]
+----
+<dependency>
+  <groupId>com.datastax.oss.quarkus</groupId>
+  <artifactId>cassandra-quarkus-client</artifactId>
+  <version>${project.version}</version>
```
absurdfarce, line 46: This isn't quite what we want, is it?  The user's project will shift versions as needed based on their development... I don't think we want to tie the version of the Quarkus extension to that.

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,301 @@
...
+
+  public Fruit(String name, String description, String id) {
+    this.name = name;
+    this.description = description;
+    this.id = id;
+  }
+  // getters, setters, hashCode and equals omitted for brevity
+}
+----
+We are using ObjectMapper, due to that fact the class is annotated with an `@Entity`.
```
absurdfarce, line 73: Perhaps "ObjectMapper" here should be a pointer to https://docs.datastax.com/en/developer/java-driver/4.6/manual/mapper/ or other relevant docs to explain what we mean?

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,301 @@
...
+----
+
+The list and add operations are executed for the `STORE_NAME` id. This is a partition key of the data model.
+We can easily retrieve all rows from cassandra using the partition key. They will be sorted by the clustering key.
+The `FruitResource` is using the `FruitService` that encapsulates the data access logic.
+
+When creating the REST API we should not share the same entity object between REST API and data access layers.
+They should not be coupled to allow evolving API independent of the storage layer.
+This is the reason why the API is using the `FruitDto`.
+This class will be used by the quarkus to convert JSON to java object for client request and java object to JSON for the response.
```
absurdfarce, line 162: Nit: s/used by the quarkus/used by Quarkus/

This translation is actually provided by REST Easy (https://resteasy.github.io/) so we might just want to say that here

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,301 @@
...
+----
+
+You can configure all java-driver settings using the `application.conf` or `application.json` files.
+They will be passed automatically to the underlying java-driver config loader.
+The settings from `application.properties` with `quarkus.cassandra` prefix have the priority.
+
+To see the full list of settings, please refer to the {reference-conf-url}[reference.conf]
+
+== Running a Cassandra Database
+As by default, `CassandraClient` is configured to access a local Cassandra database on port 9042 (the default Cassandra port).
```
absurdfarce, line 210: s/As by default/By default/

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,301 @@
...
+   --name local-cassandra-instance \
+   -p 7000:7000 \
+   -p 7001:7001 \
+   -p 7199:7199 \
+   -p 9042:9042 \
+   -p 9160:9160 \
+   -p 9404:9404 \
+   -d \
+   launcher.gcr.io/google/cassandra3
+----
```
absurdfarce, line 227: Along the same lines as what we did in the quickstarter doc we should mention here that 9042 is required, all others are optional (but provide enhanced features like JMX monitoring of the C* instance etc.)

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,301 @@
...
+[source,shell]
+----
+docker exec -it local-cassandra-instance cqlsh -e "CREATE KEYSPACE IF NOT EXISTS k1 WITH replication = {'class':'SimpleStrategy', 'replication_factor':1}"
+----
+
+And the Cassandra table that will be used by the quickstart application.
+[source,shell]
+----
+docker exec -it local-cassandra-instance cqlsh -e "CREATE TABLE IF NOT EXISTS k1.fruit(id text, name text, description text, PRIMARY KEY((id), name))"
+----
```
absurdfarce, line 239: I wonder if it might read more cleanly to combine these into a single paragraph and show both code samples after the fact:

"Next you need to create the keyspace and table that will be used by your application:"
<sample1>
<sample2>

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,301 @@
...
+   launcher.gcr.io/google/cassandra3
+----
+
+Next, you need to create the keyspace:
+[source,shell]
+----
+docker exec -it local-cassandra-instance cqlsh -e "CREATE KEYSPACE IF NOT EXISTS k1 WITH replication = {'class':'SimpleStrategy', 'replication_factor':1}"
+----
+
+And the Cassandra table that will be used by the quickstart application.
```
absurdfarce, line 235: This should probably be "your sample application" rather than "the quickstart application" here

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,301 @@
...
+[source,shell]
+----
+docker exec -it local-cassandra-instance cqlsh -e "CREATE TABLE IF NOT EXISTS k1.fruit(id text, name text, description text, PRIMARY KEY((id), name))"
+----
+
+For the local Cassandra that is not running within docker, you need to execute cqlsh commands:
```
absurdfarce, line 241: Perhaps something like "If you're running Cassandra locally you can execute the cqlsh commands directly:" might be cleaner here

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,301 @@
...
+
+So when you access the `/metrics` endpoint of your application you will have information about added metrics.
+
+== Building a native executable
+
+You can use the Cassandra client in a native executable.
+
+You can build a native executable with the `mvn clean package -Pnative` command.
+
+Running it is as simple as executing `./target/quickstart-1.0.0-SNAPSHOT-runner`.
```
absurdfarce, line 295: We've been describing this as creating a new app which happens to have the same content as the quickstart app rather than using the quickstart app itself.  Perhaps we should just tell the user to look for an app ending in "-runner" in the "target" directory here?

tomekl007:
> The JIRA ticket mentioned a CONTRIBUTING doc for this repo as well... is that still something we plan on adding?

It was moved to a new ticket:
https://datastax-oss.atlassian.net/browse/JAVA-2761

tomekl007:
> It's interesting that most of the Quarkus extensions (or at least those defined at https://github.com/quarkusio/quarkus/tree/master/extensions) don't have per-extension documentation. 

I would say that most projects have the documentation, see:
https://quarkus.io/guides/

>It _looks like_ their docs are just managed as part of the main project doc at https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc. Is this still an ongoing thing? Should we be making an effort to get our documentation into src/main/asciidoc rather than within the extension itself?
We would have it on their repo if our extension will be in their repo. We have a separate repository, that is why disc are not there. They want to fetch the docs later, but there is no process for doing it yet.

tomekl007:
> Note: asciidoc used here as it's the primary documentation method for Quarkus... see https://github.com/quarkusio/quarkus/blob/master/docs/src/main/asciidoc/README.adoc

I don't understand this comment - I've created all docs in the asciidoc format as well, so where's the problem?

documentation/src/main/asciidoc/cassandraclient.adoc
tomekl007: It is split now already:
```

Next, you need to create the keyspace:
[source,shell]
----
docker exec -it local-cassandra-instance cqlsh -e "CREATE KEYSPACE IF NOT EXISTS k1 WITH replication = {'class':'SimpleStrategy', 'replication_factor':1}"
----

And the Cassandra table that will be used by the quickstart application.
[source,shell]
----
docker exec -it local-cassandra-instance cqlsh -e "CREATE TABLE IF NOT EXISTS k1.fruit(id text, name text, description text, PRIMARY KEY((id), name))"
```

documentation/src/main/asciidoc/cassandraclient.adoc
tomekl007: When the user will follow this tutorial, he should copy the `pom.xml` from the quickstart. If he will do it, the name of the app will be quickstart - so the commands are proper

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,303 @@
...
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public class FruitResource {
+
+  private static final String STORE_NAME = "acme";
+  @Inject FruitService fruitService;
+
+  @GET
+  public List<Fruit> list() {
+    return fruitService.get(STORE_NAME);
```
adutra, line 141: I definitely don't understand this data model. Isn't `id` supposed to be the fruit id?

documentation/src/main/asciidoc/cassandraclient.adoc
adutra: He was asking the opposite: that you combine the snippets together. I think I agree it would read more cleanly.

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,303 @@
...
+docker exec -it local-cassandra-instance cqlsh -e "CREATE KEYSPACE IF NOT EXISTS k1 WITH replication = {'class':'SimpleStrategy', 'replication_factor':1}"
+----
+
+And the Cassandra table that will be used by your quickstart application.
+[source,shell]
+----
+docker exec -it local-cassandra-instance cqlsh -e "CREATE TABLE IF NOT EXISTS k1.fruit(id text, name text, description text, PRIMARY KEY((id), name))"
+----
+
+For the local Cassandra that is not running within docker, you need to execute cqlsh commands:
```
adutra, line 243:
```suggestion
If you're running Cassandra locally you can execute the cqlsh commands directly:
```

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,303 @@
...
+
+This behavior can be disabled by setting the `quarkus.cassandra.health.enabled` property to `false` in your `application.properties`.
+
+== Metrics
+
+If you are using the `quarkus-smallrye-metrics` extension, `cassandra-quarkus` can provide metrics about CqlSession and Cassandra nodes.
+This behavior must first be enabled by setting the `quarkus.cassandra.metrics.enabled` property to `true` in your `application.properties`.
+
+The next step that you need to do is set explicitly which metrics should be enabled.
+The `quarkus.cassandra.metrics.session-enabled` and `quarkus.cassandra.metrics.node-enabled` should be used for enabling metrics.
```
adutra, line 277:
```suggestion
The `quarkus.cassandra.metrics.session-enabled` and `quarkus.cassandra.metrics.node-enabled` properties should be used for enabling metrics; the former should contain a list of session-level metrics to enable, while the latter should contain a list of node-level metrics to enable. Both properties accept a comma-separated list of valid metric names.
```

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,303 @@
...
+So for example to enable `session.connected-nodes` and `node.pool.open-connections` you should add the following settings to your
+`application.properties`:
+[source, properties]
+----
+quarkus.cassandra.metrics.enabled=true
+quarkus.cassandra.metrics.session-enabled=connected-nodes
```
adutra, line 283: What's the exact syntax to use if there is more than one metric enabled? Maybe you should add an example of this.

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,303 @@
...
+quarkus.cassandra.metrics.node-enabled=pool.open-connections
+----
+
+For the full list of available metrics, please refer to the {reference-conf-url}[reference.conf] and the `advanced.metrics` section.
+
+So when you access the `/metrics` endpoint of your application you will have information about added metrics.
```
adutra, line 289:
```suggestion
When metrics are properly enabled and when you access the `/metrics` endpoint of your application, you will see metric reports for all enabled metrics.
```

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,303 @@
...
+
+You can build a native executable with the `mvn clean package -Pnative` command.
+
+Running it is as simple as executing `./target/quickstart-1.0.0-SNAPSHOT-runner`.
+
+You can then point your browser to `http://localhost:8080/fruits.html` and use your application.
+
+== Conclusion
+
+Accessing a Cassandra database from a Cassandra Client is easy with Quarkus as it provides configuration and native support for it.
```
adutra, line 303:
```suggestion
Accessing a Cassandra database from a client application is easy with Quarkus and the Cassandra extension, which provides configuration and native support for the DataStax Java driver for Apache Cassandra.
```

documentation/src/main/resources/theme/default-theme.yml
```diff
@@ -0,0 +1,287 @@
+font:
```
adutra, line 1: Is this imposed by Quarkus? Shouldn't this be imported from some URL rather than stored in-tree?

documentation/src/main/resources/theme/fonts/LICENSE-Overpass.txt
```diff
@@ -0,0 +1,92 @@
+Copyright (c) 2016 by Red Hat, Inc. All rights reserved.
```
adutra, line 1: This worries me, is this license compatible with Apache license 2? We we really need to use this font?

documentation/src/main/resources/theme/quarkus-theme.yml
```diff
@@ -0,0 +1,314 @@
+# Hibernate theme based on the default theme
```
adutra, line 1: Same here, how about importing this file from some external location?

documentation/src/main/asciidoc/cassandraclient.adoc
tomekl007: I was referring to DataStax Object Mapper, link to docs added.

documentation/src/main/asciidoc/cassandraclient.adoc
tomekl007: done

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,303 @@
...
+== Creating JSON REST service
+In this example, we will create an application to manage a list of fruits.
+
+First, let's create the `Fruit` bean as follows:
+[source,java]
+----
+@Entity
+public class Fruit {
+
+  @ClusteringColumn private String name;
```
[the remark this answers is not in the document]
tomekl007: My idea was that `id` is a `store_name`:
`@PartitionKey private String id;`
and the REST API is inserting and retrieving data only for one particular store name. It is solving problem of fetch all. I can change the id column to `store_name`, wdyt?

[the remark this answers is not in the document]
adutra: But you need to change the quickstart project too, right? I agree that `store_name` is better, but since it's not related to this PR, how about doing it in a separate PR?

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,303 @@
...
+Now create a `com.datastax.oss.quarkus.FruitService` that will be the business layer of our application and store/load the fruits from the Cassandra database.
+[source,java]
+----
+@ApplicationScoped
+public class FruitService {
+  private final FruitDao dao;
+
+  @Inject
+  public FruitService(CqlSession session) {
+    dao = new FruitMapperBuilder(session).build().fruitDao(CqlIdentifier.fromCql("k1"));
```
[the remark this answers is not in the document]
tomekl007: 1. I can create an application bean that is `FruitDao`, but I am not sure if it is needed. It will add some complexity to this example by adding one additional layer, wdyt?

2. Regarding using the session name from the config file, I am not sure if we should provide it.
- We can add one additional setting in the application.properties of our starter called keyspace, but it will restrict usage of the `CqlSession` to one keyspace. I think it may be too limiting.
- Right now, the problem of providing keyspace name is on the user - he needs to use any mechanism that he wants, to propagate it. The user when creating own quarkus app can use their own `@ConfigRoot` and propagate it, or have it hardcoded. 

I am not sure which direction is better, for spring we did configure keyspace at the start and restricted the `CqlSession` to one keyspace. Maybe we should wait for feedback before changing it?

documentation/src/main/asciidoc/cassandraclient.adoc
tomekl007: It is listing all fruits withing STORE_ID partition key.

documentation/src/main/asciidoc/cassandraclient.adoc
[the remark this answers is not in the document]
tomekl007: Ok, I can do followup PR with that change.
I've created a JIRA task to not forget about it:
https://datastax-oss.atlassian.net/browse/JAVA-2767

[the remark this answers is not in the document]
adutra: Thank you!!

documentation/src/main/resources/theme/default-theme.yml
tomekl007: That's a problem because we are the first ones (similar to Netty EvenGroup) so I doubt that there is a way to do it.
I did ask a question in the quarkus ticket:
https://github.com/quarkusio/quarkus/issues/7876#issuecomment-626725515

documentation/pom.xml
```diff
@@ -0,0 +1,192 @@
...
+              <tarLongFileMode>posix</tarLongFileMode>
+            </configuration>
+          </execution>
+        </executions>
+      </plugin>
+    </plugins>
+  </build>
+  <profiles>
+    <profile>
+      <id>documentation-pdf</id>
```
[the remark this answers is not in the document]
tomekl007: I think that the overhead of generating docs is very small so maybe we could simplify and just generate HTML + PDF on every build, wdyt?

documentation/src/main/resources/theme/fonts/LICENSE-Overpass.txt
tomekl007: https://github.com/datastax/cassandra-quarkus/pull/9#discussion_r423068194

documentation/src/main/resources/theme/quarkus-theme.yml
tomekl007: https://github.com/datastax/cassandra-quarkus/pull/9#discussion_r423068194

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,299 @@
+= Quarkus - Using the Cassandra Client
+
+Cassandra is a well known NoSQL Database that is widely used.
```
olim7t, line 3: This is a bit vague. Even if this is just a brief introductory description, we could be more specific. The first sentence of the Wikipedia article sounds like a good starting point:

> Apache Cassandra is a free and open-source, distributed, wide column store, NoSQL database management system designed to handle large amounts of data across many commodity servers, providing high availability with no single point of failure.

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,299 @@
...
+However, you can go right to the completed example.
+
+The solution is located in the `quickstart` {quickstart-clone-url}[directory].
+
+== Creating the Maven project
+
+First, we need a new project. Create a new maven project and copy the `pom.xml` file that is present in the `quickstart` directory.
+
+The `pom.xml` is importing the RESTEasy/JAX-RS, JSON-B, Context Propagation and Cassandra Client extensions.
+We will be building a REST application using the DataStax Object Mapper {mapper-docs-url}[DataStax Object Mapper].
```
olim7t, line 38:
```suggestion
We will be building a REST application using the {mapper-docs-url}[DataStax Object Mapper].
```

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,299 @@
...
+
+The most important part of the `pom.xml` is adding the `cassandra-quarkus` extension:
+[source,xml]
+----
+<dependency>
+  <groupId>com.datastax.oss.quarkus</groupId>
+  <artifactId>cassandra-quarkus-client</artifactId>
+  <version>${quarkus.version}</version>
+</dependency>
+----
```
olim7t, line 49: Reminder to maybe update this if we decide to switch to the BOM (per other private discussion).

documentation/src/main/asciidoc/cassandraclient.adoc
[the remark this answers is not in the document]
olim7t: Keyspace: right now it's hardcoded, so it's even less flexible. I think moving it to the configuration is good enough for this example. The only case it doesn't cover is having two `FruitService` beans that hit different keyspaces: that's possible, but doesn't seem like a very common use case.

Injecting `FruitDao`: I'm not sure if I would go that far, but we should at least inject the `FruitMapper`. A typical application can probably have more services (`StoreService`, `UserService`, etc), and in such a case you don't want to re-create a mapper instance for every service.

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,299 @@
...
+In this example, we are using a single instance running on localhost:
+
+[source,properties]
+----
+# configure the Cassandra client for a single instance on localhost
+quarkus.cassandra.contact-points=127.0.0.1:9042
+quarkus.cassandra.load-balancing-policy.local-datacenter=datacenter1
+----
+
+You can configure all Java driver settings using `application.conf` or `application.json` files.
```
olim7t, line 205: Maybe specify where these files should be located.

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,299 @@
...
+They will be passed automatically to the underlying driver configuration mechanism.
+Settings defined in `application.properties` with the `quarkus.cassandra` prefix will have priority over settings defined in `application.conf` or `application.json`.
+
+To see the full list of settings, please refer to the {reference-conf-url}[reference.conf].
+
+== Running a Cassandra Database
+By default, `CassandraClient` is configured to access a local Cassandra database on port 9042 (the default Cassandra port).
+If you have a local running database on this port, you should assert that:
+`quarkus.cassandra.load-balancing-policy.local-datacenter` setting is equal to dc of your local cassandra instance.
```
olim7t, line 214: Maybe we should specify that the value can be found with `SELECT data_center FROM system.local`, that might not be obvious for users not familiar with Cassandra.

documentation/src/main/asciidoc/cassandraclient.adoc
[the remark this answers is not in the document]
tomekl007: 1. Keyspace from settings implemented in https://github.com/datastax/cassandra-quarkus/pull/9/commits/97821a137d3fee2823388eef83979271e242e30a and docs 
https://github.com/datastax/cassandra-quarkus/pull/9/commits/76a99f9f0f2ecb945b50a28ac70af674f3e606a3

2. Injecting `FrutiMapper`: implemented in https://github.com/datastax/cassandra-quarkus/pull/9/commits/ab57a565a7d5fd85e82d81a41a39c19d20b4720b and docs
https://github.com/datastax/cassandra-quarkus/pull/9/commits/cb0be06dc99f4762bcbaee47da3e9337deb9ae0e

documentation/src/main/asciidoc/cassandraclient.adoc
tomekl007: Getting back to this discussion, I am hesitating about adding it. 
By using our bom, we will get rid of only one dependency, that is:
```
  <dependency>
      <groupId>com.datastax.oss.quarkus</groupId>
      <artifactId>cassandra-quarkus-client</artifactId>
      <version>${project.version}</version>
    </dependency>
```
by replacing it with:
```
<dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>io.quarkus</groupId>
        <artifactId>cassandra-quarkus-bom</artifactId>
        <version>${project.version}</version>
        <type>pom</type>
        <scope>import</scope>
      </dependency>
    </dependencies>
  </dependencyManagement>
```
I am not sure if that is a good solution. I imagine that not all users are familiar with boms. Also, other quick start guides are using direct dependency to extension, instead of the bom (they are using bom internally in the extension as well)

adutra: Frankly I think these boms are a bit overkill. They contain 3/4 artifacts each. That's not enough to justify their presence.

documentation/assembly-pdf.xml
```diff
@@ -0,0 +1,27 @@
...
+  </formats>
+  <includeBaseDirectory>true</includeBaseDirectory>
+  <fileSets>
+    <fileSet>
+      <directory>target/generated-docs-pdf</directory>
+      <includes>
+        <include>*.pdf</include>
+      </includes>
+      <excludes>
+        <exclude>0-glossary.pdf</exclude>
```
adutra, line 15: Why are these files being generated if we don't need them?

documentation/src/main/resources/theme/default-theme.yml
adutra: Ok I think it's acceptable for now. Maybe we can open a follow-up Jira ticket and repeat your question there.

documentation/assembly-pdf.xml
tomekl007: That was left-over from copy-paste, sorry. Removed

documentation/src/main/resources/theme/default-theme.yml
tomekl007: Ok, created https://datastax-oss.atlassian.net/browse/JAVA-2771

documentation/pom.xml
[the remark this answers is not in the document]
tomekl007: Implemented in https://github.com/datastax/cassandra-quarkus/pull/9/commits/62c4355fb5f42b38bf4cd3ce129b42773a486122

documentation/pom.xml
```diff
@@ -34,102 +34,98 @@
...
+            </executions>
+            <dependencies>
+              <dependency>
+                <groupId>org.asciidoctor</groupId>
+                <artifactId>asciidoctorj</artifactId>
+                <version>${asciidoctorj.version}</version>
+              </dependency>
+            </dependencies>
+          </plugin>
           <plugin>
```
adutra, line 129: You are repeating the plugin declaration twice. You should instead merge the 2 executions into one single plugin declaration.

documentation/pom.xml
```diff
@@ -34,102 +34,98 @@
...
       <activation>
         <property>
           <name>documentation-pdf</name>
         </property>
       </activation>
       <build>
         <plugins>
+          <plugin>
+            <groupId>org.apache.maven.plugins</groupId>
+            <artifactId>maven-assembly-plugin</artifactId>
```
adutra, line 49: Same here, duplicated plugin declaration.

documentation/pom.xml
```diff
@@ -34,102 +34,98 @@
...
+            <version>3.1.0</version>
+            <executions>
+              <execution>
+                <id>assemble</id>
```
adutra, line 53: Give a more meaningful name to this execution: what is it assembling exactly? Maybe `assemble-pdf`.

documentation/assembly-pdf.xml
```diff
@@ -11,16 +11,6 @@
       <includes>
         <include>*.pdf</include>
```
adutra, line 12: Now that I think of it: how many files are we including, just 1? Maybe we don't need this assembly at all?

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,347 @@
...
+----
+
+The `list` and `add` operations are executed for the `STORE_NAME` id. This is a partition key of the data model.
+We can easily retrieve all rows from cassandra using the partition key. They will be sorted by the clustering key.
+`FruitResource` is using `FruitService` which encapsulates the data access logic.
+
+When creating the REST API we should not share the same entity object between REST API and data access layers.
+They should not be coupled to allow the API to evolve independently of the storage layer.
+This is the reason why the API is using a `FruitDto` class.
+This class will be used by Quarkus to convert JSON to java object for client request and java object to JSON for the response.
```
absurdfarce, line 209: I agree that this is the right decision architecturally... but it's not what we're doing right now.

More specifically we are doing the right thing in the POST handler; we're expecting a JSON rep of a FruitDto and working with it as such.  But the handler for the GET makes no use of FruitDto at all.  And there's nothing in Quarkus or REST Easy that knows that we should render Fruit as a FruitDto instead (or knows how to translate one into the other).

To validate, note what we get with a simple curl request to the Web service:

```
[mersault@localhost ~]$ curl http://localhost:8080/fruits
[{"description":"foo","id":"acme","name":"bar"},{"description":"bar","id":"acme","name":"foo"}]
```

The "id" field is not in FruitDto... but it is in Fruit.  What's actually happening here is that the collection of Fruit objects returned from FruitResource.list() is being serialized to JSON directly.  But, as the comment suggests, we should be using FruitDto for this operation.

We need something like the following instead:

```java
  @GET
  public List<FruitDto> list() {
    return fruitService.get(STORE_NAME).stream()
            .map(fruit -> new FruitDto(fruit.getName(), fruit.getDescription()))
            .collect(Collectors.toList());
  }
```

Note the output of the curl request with this change:

```
[mersault@localhost ~]$ curl http://localhost:8080/fruits
[{"description":"foo","name":"bar"},{"description":"bar","name":"foo"}]
```

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,347 @@
...
+They need to be located in the resources directory of your application.
+They will be passed automatically to the underlying driver configuration mechanism.
+Settings defined in `application.properties` with the `quarkus.cassandra` prefix will have priority over settings defined in `application.conf` or `application.json`.
+
+To see the full list of settings, please refer to the {reference-conf-url}[reference.conf].
+
+== Running a Cassandra Database
+By default, `CassandraClient` is configured to access a local Cassandra database on port 9042 (the default Cassandra port).
+If you have a local running database on this port, you should assert that:
+`quarkus.cassandra.load-balancing-policy.local-datacenter` setting is equal to dc of your local cassandra instance.
```
absurdfarce, line 261: Nit: "setting is equal to the DC of your local Cassandra instance".  We capitalized "Cassandra" everywhere else in this paragraph so we probably should here.

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,347 @@
...
+docker exec -it local-cassandra-instance cqlsh -e "CREATE KEYSPACE IF NOT EXISTS k1 WITH replication = {'class':'SimpleStrategy', 'replication_factor':1}"
+docker exec -it local-cassandra-instance cqlsh -e "CREATE TABLE IF NOT EXISTS k1.fruit(id text, name text, description text, PRIMARY KEY((id), name))"
+----
+
+If you're running Cassandra locally you can execute the cqlsh commands directly:
+[source,shell]
+----
+cqlsh -e "CREATE KEYSPACE IF NOT EXISTS k1 WITH replication = {'class':'SimpleStrategy', 'replication_factor':1}
+cqlsh -e "CREATE TABLE IF NOT EXISTS k1.fruit(id text, name text, description text, PRIMARY KEY((id), name))
+----
```
absurdfarce, line 292: Here we _can_ point people to ```cqlsh -f setup_db.cql```, right?  Might be simpler than having to show the statements directly inline again.

documentation/pom.xml
tomekl007: We already have `<id>assemble-pdf</id>` phase

documentation/README.md
```diff
@@ -0,0 +1,17 @@
+# Generating Documentation
+
+To generate the documentation run:
+
+    mvn package -Prelease
```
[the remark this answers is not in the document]
tomekl007: I am not getting this warning now after removing duplicated plugin

documentation/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+## HTML docs
+To see the documentation in the HTML format:
+
+    open target/generated-docs/cassandraclient.html
```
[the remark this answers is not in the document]
tomekl007: Please run it once again. It is generating for me

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,347 @@
...
+
+The elements are stored in the Cassandra Database.
+
+
+== Solution
+
+We recommend that you follow the instructions in the next sections and create the application step by step.
+However, you can go right to the completed example.
+
+The solution is located in the `quickstart` {quickstart-clone-url}[directory].
```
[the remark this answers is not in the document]
tomekl007: Screenshot 2020-05-14 at 10 28 11
It is generating for me

documentation/src/main/asciidoc/cassandraclient.adoc
tomekl007: I disagree. It is simpler to have those statements in the docs. Otherwise, the user will need to lookup another file.

documentation/assembly-pdf.xml
tomekl007: For pdf, there is one file. For HTML version there are 3. We can remove the assembly for pdf, but don't you think that it is better to have the common mechanism for distributing docs for HTML and pdf?

adutra: I'm fine either way.

documentation/src/main/asciidoc/cassandraclient.adoc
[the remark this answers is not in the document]
adutra: It is rendered correctly for me too.

documentation/README.md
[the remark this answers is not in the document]
absurdfarce: Confirmed, it seems to be generating now.  Weird.

I am now seeing some warnings which _appear_ to come from the run of asciidoc:

```
[INFO] --- asciidoctor-maven-plugin:2.0.0-RC.1:process-asciidoc (output-pdf) @ cassandra-quarkus-documentation ---
[INFO] Using 'UTF-8' encoding to copy filtered resources.                                
[INFO] Copying 0 resource                                                                                                                                                        
uri:classloader:/gems/prawn-svg-0.30.0/lib/prawn/svg/color.rb:212: warning: shadowing outer local variable - result
uri:classloader:/gems/prawn-svg-0.30.0/lib/prawn/svg/properties.rb:51: warning: `*' interpreted as argument prefix                                                               
uri:classloader:/gems/treetop-1.6.10/lib/treetop/compiler/grammar_compiler.rb:37: warning: shadowing outer local variable - ext
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized                                                                    
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized                                                                    
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized                                                                    
uri:classloader:/gems/ttfunk-1.5.1/lib/ttfunk/table.rb:26: warning: instance variable @offset not initialized
uri:classloader:/gems/prawn-2.2.2/lib/prawn/font/ttf.rb:118: warning: instance variable @italic_angle not initialized                                                            
uri:classloader:/gems/prawn-2.2.2/lib/prawn/font/ttf.rb:118: warning: instance variable @italic_angle not initialized
uri:classloader:/gems/prawn-2.2.2/lib/prawn/font/ttf.rb:118: warning: instance variable @italic_angle not initialized                                                            
[INFO] Rendered /work/git/cassandra-quarkus/documentation/src/main/asciidoc/cassandraclient.adoc
```

I don't recall seeing those before.

documentation/src/main/asciidoc/cassandraclient.adoc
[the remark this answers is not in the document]
absurdfarce: It is now for me as well too... between this and the "only PDF" comment above I'm starting to wonder what was up with my runs of this thing under the prior commits.  I ran it multiple times (with cleans in between) and saw consistent results... but I agree this is fixed now.

documentation/src/main/asciidoc/cassandraclient.adoc
absurdfarce: I"m not sure I agree, but I don't feel terribly strongly about it... so let's keep them as they are.

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,349 @@
...
+They need to be located in the resources directory of your application.
+They will be passed automatically to the underlying driver configuration mechanism.
+Settings defined in `application.properties` with the `quarkus.cassandra` prefix will have priority over settings defined in `application.conf` or `application.json`.
+
+To see the full list of settings, please refer to the {reference-conf-url}[reference.conf].
+
+== Running a Cassandra Database
+By default, `CassandraClient` is configured to access a local Cassandra database on port 9042 (the default Cassandra port).
+If you have a local running database on this port, you should assert that:
+`quarkus.cassandra.load-balancing-policy.local-datacenter` setting is equal to dc of your local Cassandra instance.
```
absurdfarce, line 263: Nit: since the name "datacenter" is used in the config probably prefer "equal to the datacenter of your local Cassandra instance" here.

documentation/src/main/asciidoc/cassandraclient.adoc
```diff
@@ -0,0 +1,349 @@
...
+   -p 7000:7000 \
+   -p 7001:7001 \
+   -p 7199:7199 \
+   -p 9042:9042 \
+   -p 9160:9160 \
+   -p 9404:9404 \
+   -d \
+   launcher.gcr.io/google/cassandra3
+----
+Note that only the 9042 port is required. All others all optional but provide enhanced features like JMX monitoring of the C* instance.
```
absurdfarce, line 280: Semi-nit: we've used the actual name "Cassandra" everywhere else, probably want that here instead of "C*"

documentation/pom.xml
adutra: These are execution ids, not phases, but I get your point, I mixed the two executions. Let's call this one `assemble-html` then.

documentation/src/main/asciidoc/cassandraclient.adoc
adutra: Leaving as is for now.

adutra:
Please open follow-up tickets if something was missing.

PR #10, opened 2020-05-08
JAVA-2745: Repo-side support for necessary workflows

absurdfarce:


.github/workflows/quarkus-snapshot.yaml
```diff
@@ -0,0 +1,51 @@
+name: "Quarkus ecosystem CI"
```
tomekl007, line 1: Why do you want to add this file to our repo? Quarkus project will not fetch our entire repository. They only want to take docs and quickstart. 
Besides that, we need to integrate with `quarkus-platform` (https://datastax-oss.atlassian.net/browse/JAVA-2708)
by creating a PR with a couple of our IT tests to their repo.

I think that in this ticket, we should take a look at the quarkus CI process and adapt our Jenkinsfile accordingly. wdyt?

absurdfarce: This file is referenced as item 4 in their README for adding a new extension to their CI process... see the fourth item at https://github.com/quarkusio/quarkus-ecosystem-ci.  It appears that this doc is necessary to define the build process that should be used to build and test the extension against whatever is providing Quarkus master.

I explicitly avoided doing anything in this PR related to Quarkus Platform since I assumed that would be handled in JAVA-2708.  If we keep distinct tickets for those goals (which seems to make sense) I'd argue to avoid doing anything related to Quarkus Platform here.

olim7t:
It took me a little while to figure out how the Quarkus version would match the one declared in our POM. If anyone else is wondering:
* [quarkus-ecosystem-ci/cassandra-quarkus/context.yaml](https://github.com/quarkusio/quarkus-ecosystem-ci/blob/master/cassandra-quarkus/context.yaml) defines the version. We use 999-SNAPSHOT, which is apparently a special number for the head of Quarkus master.
* [quarkus-ecosystem-ci/setup-and-test](https://github.com/quarkusio/quarkus-ecosystem-ci/blob/master/setup-and-test) checks out Quarkus and installs it locally. It looks for a `quarkus.sha` variable in `context.yaml` but we don't define it, so I'm not sure what happens here (maybe it gets injected automatically, or maybe it's just empty and `git checkout` is a no-op?)
*  [quarkus-ecosystem-ci/quarkus-ecosystem-test](https://github.com/quarkusio/quarkus-ecosystem-ci/blob/master/quarkus-ecosystem-test) runs our build, but it forces the `quarkus.version` system property to what was declared in `context.yaml`. So this relies on the fact that our POM has a `quarkus.version` property, this is important to know and I think we should document it in the POM.

olim7t:
> maybe it gets injected by some other script, or maybe it's just empty and git checkout is a no-op?

It looks like it's injected automatically, see [quarkus-ecosystem-ci/.github/workflows/ecosystem-action.yml](https://github.com/quarkusio/quarkus-ecosystem-ci/blob/master/.github/workflows/ecosystem-action.yml#L29). I'm still missing a piece because it looks like that script commits the change, but I don't see a `quarkus.sha` in the commits for already present extensions.

.github/workflows/quarkus-snapshot.yaml
```diff
@@ -0,0 +1,51 @@
...
+env:
+  ECOSYSTEM_CI_REPO: quarkusio/quarkus-ecosystem-ci
+  ECOSYSTEM_CI_REPO_FILE: context.yaml
+  JAVA_VERSION: 11
+
+  #########################
+  # Repo specific setting #
+  #########################
+
+  ECOSYSTEM_CI_REPO_PATH: FIXME # TODO: this needs to be set to the directory name added in 'quarkusio/quarkus-ecosystem-ci'
```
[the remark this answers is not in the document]
absurdfarce: Yeah, this was just an outright misunderstanding on my part.  Fixed now.

absurdfarce:
To clarify comments from @olim7t above: context.yaml is actually created on their side as part of the commit op... see https://github.com/quarkusio/quarkus-ecosystem-ci/blob/master/.github/workflows/ecosystem-action.yml#L42-L48.  That's where the "999-SNAPSHOT" bit comes from.  QUARKUS_SHA is extracted from "git ls-remote" in ecosystem-action.yml so it should always be current as well.  These values are then written to context.yaml which in turn is committed.  All the extension directories (including ours) in quarkus-ecosystem-ci do contain a context.yaml with a defined quarkus.version and quarkus.sha value.

Related to the above point: master for quarkusio/quarkus has a default version of "999-SNAPSHOT".  See https://github.com/quarkusio/quarkus/blob/master/pom.xml#L15

QUARKUS_VERSION is then passed through to quarkus-ecosystem-test which in turn passes it to the extension build via quarkus.version (as described by @olim7t above).  This appeared to match up to the convention we have in the POM of defining versions via props but I completely agree that explicitly documenting this interconnection is worthwhile.  I'll add a note indicating this dependency.

olim7t:
> QUARKUS_SHA is extracted from "git ls-remote" in ecosystem-action.yml so it should always be current as well.

I could have sworn that none of the extensions' `context.yaml` had the SHA when I checked last night, but indeed it's there now.

absurdfarce:
Oh, and yq is apparently an attempt "to be the jq or sed of yaml files".  See https://github.com/mikefarah/yq

pom.xml
```diff
@@ -27,6 +27,7 @@
   <url>https://quarkus.io/guides/cassandra</url>
   <inceptionYear>2020</inceptionYear>
   <properties>
+    <!-- Note: this property is used by the Quarkus CI to pass version info into the build.  See JAVA-2745 for details -->
```
olim7t, line 30: 👍

PR #12, opened 2020-05-14
JAVA-2775: Fixes for quarkus-ecosystem-ci test failures

absurdfarce:


absurdfarce:
I've validated that the specific test failure defined in the ticket passes against both master (as of the specified commit) and 1.3.0-Final (the base version we have in the extension).  So at least we haven't done any damage here.

PR #13, opened 2020-05-16
JAVA-2776: Integration test fix for test failures on quarkus-ecosystem-ci

absurdfarce:


absurdfarce:
I've run the test suite + this change against a local checkout of Quarkus master as well as 1.3.0 and all tests pass in both cases.

integration-tests/src/test/java/com/datastax/oss/quarkus/CassandraHealthCheckIT.java
```diff
@@ -15,36 +15,27 @@
...
-    assertThat(checks.size()).isOne();
-    assertThat(checks.get(0).get("name"))
-        .isEqualTo("DataStax Apache Cassandra Driver health check");
+    when()
+        .get("/health/ready")
+        .then()
+        .statusCode(Status.OK.getStatusCode())
+        .body("status", equalTo("UP"))
+        .body("checks", hasSize(1))
+        .body("checks[0].name", equalTo("DataStax Apache Cassandra Driver health check"));
```
absurdfarce, line 39: Re-wrote this test (and the other) to better leverage the rest-assured DSL.  It's now more declarative, which in turn should make what it's testing that much clearer.  The DSL explicitly exposes the Hamcrest matchers as part of it's API so this did strongly suggest converting to use those matchers, which I've done here.  I'm well aware that we use assertj in driver code but for a Quarkus integration (more specifically a Quarkus integration for which the hamcrest lib is a first-class citizen and part of the testing API) this seemed like a reasonable change.

integration-tests/pom.xml
```diff
@@ -107,6 +107,7 @@
     <dependency>
       <groupId>io.rest-assured</groupId>
       <artifactId>rest-assured</artifactId>
+      <version>4.2.0</version>
```
adutra, line 110: Quarkus BOM seems to be importing that version already, so I'm not sure we need to explicitly declare the version here.

integration-tests/src/test/java/com/datastax/oss/quarkus/CassandraHealthCheckIT.java
```diff
@@ -15,36 +15,27 @@
  */
 package com.datastax.oss.quarkus;
 
-import static org.assertj.core.api.Assertions.assertThat;
+import static io.restassured.RestAssured.*;
```
adutra, line 18: We don't have proper contribution guidelines for this project yet, but in the meanwhile I suggest we follow the ones of the driver, which forbids star imports.

integration-tests/src/test/java/com/datastax/oss/quarkus/CassandraHealthCheckIT.java
adutra: Sounds very reasonable, and definitely easier to read. 👍

integration-tests/pom.xml
adutra: Nevermind I read your explanations in the Jira ticket.

absurdfarce:
Based on your review it seems like the star imports were the only remaining concern @adutra ... let me know if I missed something else.

PR #14, opened 2020-05-19
JAVA-2782: Add substitutions to make Guava more Graal-friendly

absurdfarce:


runtime/src/main/java/com/datastax/oss/driver/shaded/guava/common/primitives/UnsignedBytesSubstitutions.java
```diff
@@ -0,0 +1,40 @@
...
+final class LexicographicalComparatorHolderSubstitutions {
+
+  @Alias
+  @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.FromAlias)
+  static Comparator<byte[]> BEST_COMPARATOR = UnsignedBytes.lexicographicalComparatorJavaImpl();
+
+  @Substitute
+  static Comparator<byte[]> getBestComparator() {
+    return UnsignedBytes.lexicographicalComparatorJavaImpl();
+  }
```
absurdfarce, line 39: There are two discrete ways to get an instance of an UnsafeComparator.  On 64-bit architectures it's the default stored in BEST_COMPARATOR, a value which is set by calling getBestComparator().  For sake of completeness we shut off both avenues of access to the now-deleted class; in fact we _have_ to do so.  If there's a viable code path which leads to a class we've explicitly marked as deleted Graal's analysis will detect it and (correctly) complain loudly.

runtime/src/main/java/com/datastax/oss/driver/shaded/guava/common/primitives/UnsignedBytesSubstitutions.java
```diff
@@ -0,0 +1,40 @@
...
+import com.oracle.svm.core.annotate.Alias;
+import com.oracle.svm.core.annotate.Delete;
+import com.oracle.svm.core.annotate.RecomputeFieldValue;
+import com.oracle.svm.core.annotate.Substitute;
+import com.oracle.svm.core.annotate.TargetClass;
+import java.util.Comparator;
+
+@TargetClass(UnsignedBytes.LexicographicalComparatorHolder.UnsafeComparator.class)
+@Delete
+final class UnsafeComparatorSubstitutions {}
```
absurdfarce, line 27: One _could_ argue that marking this class as deleted (or more accurately "unavailable once this substitution is applied") is a bit heavy-handed.  I think it's a nice bit of insurance; with this step any code path which _could_ lead to an instance of this class will trigger build failures in Graal's analysis phase.  This is desirable; it _forces_ the more rigorous impl below.

absurdfarce:
I went back and forth on whether the substitution class should have some comments spelling out the problem.  Philosophies always vary here but it seemed like terseness was more consistent with the rest of the code base.  The problem is already stated fairly clearly on the JIRA ticket so _maybe_ a pointer to that wouldn't be the worst thing.

absurdfarce:
After adding this substitution I started seeing some odd errors indicating that Graal's analysis was finding paths to ClassLoader.defineClass() in the jnr-posix code.  This was... odd since this very behaviour was addressed in JAVA-2663.  Then I remembered that the Java driver version in use here is 4.6.1; 2663 went in _after_ that was released.  So when we update the Java driver version this will disappear.  I confirmed this expectation by testing using a 4.7.0-SNAPSHOT driver which contained 2663... as expected no warnings were displayed at all.

absurdfarce:
My output locally:

```
[INFO] [io.quarkus.deployment.pkg.steps.NativeImageBuildStep] /work/local/graalvm-ce-java8-19.3.1/jre/bin/native-image -J-Dsun.nio.ch.maxUpdateArraySize=100 -J-Djava.util.logging
.manager=org.jboss.logmanager.LogManager -J-Dvertx.logger-delegate-factory-class-name=io.quarkus.vertx.core.runtime.VertxLogDelegateFactory -J-Dvertx.disableDnsResolver=true -J-D
io.netty.leakDetection.level=DISABLED -J-Dio.netty.allocator.maxOrder=1 -J-Duser.language=en -J-Dfile.encoding=UTF-8 --initialize-at-build-time= -H:InitialCollectionPolicy=com.or
acle.svm.core.genscavenge.CollectionPolicy$BySpaceAndTime -H:+JNI -jar cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner.jar -H:FallbackThreshold=0 -H:+ReportExceptionStackTrace
s -H:-AddAllCharsets -H:-IncludeAllTimeZones -H:EnableURLProtocols=http -H:NativeLinkerOption=-no-pie --no-server -H:-UseServiceLoaderFeature -H:+StackTrace cassandra-quarkus-qui
ckstart-1.0.0-SNAPSHOT-runner                                                                                                                                                     
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]    classlist:  22,017.68 ms                                                                                            
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]        (cap):   1,375.96 ms                  
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]        setup:   4,013.92 ms                                                                                            
00:02:59,535 INFO  [com.dat.oss.dri.int.cor.DefaultMavenCoordinates] DataStax Java driver for Apache Cassandra(R) (com.datastax.oss:java-driver-core) version 4.7.0-SNAPSHOT
00:03:01,790 INFO  [org.jbo.threads] JBoss Threads version 3.0.1.Final                                                                                                            
00:03:36,847 INFO  [com.dat.oss.dri.int.cor.os.Native] Using Graal-specific native functions                                                                                     
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]   (typeflow):  82,497.52 ms                                                                                            
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]    (objects):  30,025.19 ms                                                                                            
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]   (features):   1,016.94 ms                                                                                            
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]     analysis: 116,896.98 ms                                                                                            
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]     (clinit):   1,679.41 ms                                                                                            
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]     universe:   4,793.83 ms            
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]      (parse):  11,510.92 ms                                                                          
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]     (inline):  12,972.32 ms                                                                                            
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]    (compile):  66,928.78 ms                                                                                            
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]      compile:  94,489.19 ms                                                
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]        image:   4,743.04 ms                                                                                            
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]        write:     788.25 ms                                                                                            
[cassandra-quarkus-quickstart-1.0.0-SNAPSHOT-runner:16683]      [total]: 248,373.21 ms                                                                                            
[INFO] [io.quarkus.deployment.QuarkusAugmentor] Quarkus augmentation completed in 250710ms  
```

absurdfarce:
Huh, JNR errors also now appearing on Jenkins... see https://jenkins-drivers.build.dsinternal.org/blue/organizations/jenkins/tools%2Fcassandra-quarkus/detail/java2782/1/pipeline/8 as an example.

Not sure why this would suddenly appear given the addition of a new substitution.  I'll dig into this a bit more, see if there's a simple way to fix this now knowing that the "right" fix is coming with JAVA-2663.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/graal/NativeSubstitutions.java
```diff
@@ -15,20 +15,40 @@
  */
```
absurdfarce, line 15: Changes to this class avoid the observed issues with ClassLoader.defineClass() via jnr by replacing the loading logic in the base impl of Native with this no-op version.

Note that this is only a temporary fix.  The _right_ fix for this problem will be provided by JAVA-2663, but until we get a driver version including that change we'll need something like this to remove jnr from consideration in the Graal case.

absurdfarce:
Native integration tests run cleanly for me locally, assuming that last failure was something environmental... so let's try this again

runtime/src/main/java/com/datastax/oss/driver/shaded/guava/common/primitives/UnsignedBytesSubstitutions.java
adutra: I'm fine with the principle, but shouldn't this substitution go into the driver itself?

runtime/src/main/java/com/datastax/oss/quarkus/runtime/graal/NativeSubstitutions.java
adutra: Wouldn't it be simpler to just use driver 4.7.0-SNAPSHOT for now? But otherwise I'm fine with the changes here as long as we don't forget to revert them when 2663 is available.

runtime/src/main/java/com/datastax/oss/driver/shaded/guava/common/primitives/UnsignedBytesSubstitutions.java
tomekl007: In the original code, the `getBestComparator()` method is invoked only by:
```
 static final Comparator<byte[]> BEST_COMPARATOR = getBestComparator();
```
So I think it should be enough to substitute `BEST_COMPARATOR` or `getBestComparator()` not both, right?

runtime/src/main/java/com/datastax/oss/quarkus/runtime/graal/NativeSubstitutions.java
```diff
@@ -15,20 +15,40 @@
...
+
+  @Substitute
+  public static boolean isCurrentTimeMicrosAvailable() {
+    return false;
+  }
+
+  @Substitute
+  public static long currentTimeMicros() {
+    throw new RuntimeException("Native operations are not supported on this platform");
+  }
```
tomekl007, line 48: If you are disabling native clock this way, I think you can remove this:
`configLoaderBuilder.withBoolean(DefaultDriverOption.TIMESTAMP_GENERATOR_FORCE_JAVA_CLOCK, true);`
from
`AbstractCassandraClientProducer`

runtime/src/main/java/com/datastax/oss/quarkus/runtime/graal/NativeSubstitutions.java
```diff
@@ -15,20 +15,40 @@
...
+
+  @KeepOriginal
```
tomekl007, line 50: What does this annotation mean? Why do we need to substitute this method? I think that it does not use any native method.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/graal/NativeSubstitutions.java
absurdfarce: We'll have to make _some_ change when 2663 comes in anyway; once we get to that spot the Native substitutions should come out entirely.  So since we already have to do work for that situation this seemed easier than publishing a snapshot.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/graal/NativeSubstitutions.java
absurdfarce: I agree, but I'd argue this should be done as part of the changes when we get a driver that contains JAVA-2663.  When that happens we can remove this restriction _and_ remove the NativeSubstitutions entirely.

absurdfarce: Just occurred to me: it probably makes sense to have a discrete ticket to represent the work referenced above.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/graal/NativeSubstitutions.java
absurdfarce: We needed to use a "Substitute" annotation on the class in order to completely replace the old implementation.  The problem here was in the static initialization of Native which means we need to prevent that code from running... and the only way I could see to do that was to replace the class entirely.  If we do that we need to provide an impl of getCpu() in this class, the one that replaces the original Native class.  "KeepOriginal" means "use the impl on the original class"... which is exactly what we want in this case.

runtime/src/main/java/com/datastax/oss/driver/shaded/guava/common/primitives/UnsignedBytesSubstitutions.java
absurdfarce: Hmmm, that's an _interesting_ point!  Doing so would benefit any other potential users of the Guava classes as well... I kinda like that.

Perhaps we do it here for now (to get the build fixed) _and_ in the driver and then remove this code when we get a version of the driver that contains the same thing?

runtime/src/main/java/com/datastax/oss/driver/shaded/guava/common/primitives/UnsignedBytesSubstitutions.java
absurdfarce: I think you're right @tomekl007 .  I originally tried just replacing getBestComparator() before I noticed all the field references.  I knew I had to fix the field but I forgot to check after the fact to see if there were any other users of getBestComparator().

All that said, I still kinda like the idea of covering both of them.  It feels weird to me to only fix one and still leave the other open as a way to get at the (now-deleted) UnsafeComparator... knowing the whole thing only works because nobody's calling getBestComparator() right now.  I think I'd rather keep both in there.

runtime/src/main/java/com/datastax/oss/driver/shaded/guava/common/primitives/UnsignedBytesSubstitutions.java
absurdfarce: I've confirmed that adding the substitution to the driver core will work.  I created a local 4.7.0-SNAPSHOT instance with the substitution in place and removed it elsewhere... and the native build was still clean.  I created a PR to represent what was done to the core to make that work... that's available at https://github.com/datastax/java-driver/pull/1443

Note a few things about that change:

* We have to add a new dependency to the core driver, although I'm pretty sure we should be able to add it in "provided" scope (still need to test that)
* I encountered a weird Javadoc generation error which required excluding the new substitution from the javadoc plugin.  I don't think this is a problem as this seems like an internal impl detail and not something we'd want to document necessarily... but it did strike me as strange.  More details on the PR referenced above.
* Adding the substitution to the core now splits the Guava package across two JARs which makes Felix and the bundler plugin unhappy... again, more details on the PR.

I think this idea is still quite viable, however I did want to be very explicit about the above behaviours to make sure we were okay with them.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/graal/NativeSubstitutions.java
absurdfarce: JAVA-2789 has been created to address this work

absurdfarce:
I'm gonna close this PR in favor of distinct PRs for the two issues at play here.  These issues are logically distinct and can be resolved at different times (the native stuff should be largely ready to go now) so there's no real reason to continue to tie them together.

PR #16, opened 2020-05-20
JAVA-2782: Native substitution fixes + driver (re)upgrade

absurdfarce:
Fix for the native substitutions to address the recent failures to build native images without requiring a specific driver version.

This PR contains the native substitution from https://github.com/datastax/cassandra-quarkus/pull/14

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -36,7 +36,7 @@
     <maven-surefire-plugin.version>3.0.0-M4</maven-surefire-plugin.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <datastax-java-driver.version>4.5.0</datastax-java-driver.version>
+    <datastax-java-driver.version>4.6.1</datastax-java-driver.version>
     <!--suppress UnresolvedMavenProperty -->
     <graalvmHome>${env.GRAALVM_HOME}</graalvmHome>
   </properties>
--- a/runtime/src/main/java/com/datastax/oss/quarkus/runtime/graal/NativeSubstitutions.java
+++ b/runtime/src/main/java/com/datastax/oss/quarkus/runtime/graal/NativeSubstitutions.java
@@ -15,20 +15,40 @@
  */
 package com.datastax.oss.quarkus.runtime.graal;
 
+import com.datastax.oss.driver.internal.core.os.CpuInfo;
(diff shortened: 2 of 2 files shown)


runtime/src/main/java/com/datastax/oss/quarkus/runtime/graal/NativeSubstitutions.java
```diff
@@ -15,20 +15,40 @@
...
+  @Substitute private static final CpuInfo.Cpu CPU = CpuInfo.determineCpu();
+
   @Substitute
   public static boolean isGetProcessIdAvailable() {
     return false;
   }
+
+  @Substitute
+  public static int getProcessId() {
+    throw new RuntimeException("Native operations are not supported on this platform");
```
adutra, line 37: `UnsupportedOperationException` would be slightly better, but I guess this doesn't matter, the method is never going to be called anyway, right?

absurdfarce: That's true at the moment @adutra , but we're relying on the user to check whether it's available before they make the call... so it's certainly something that could happen.

I agree that UOE is more directly applicable here, but it's worth noting that the "real" Native throws an IllegalStateException in this case (see https://github.com/datastax/java-driver/blob/f4ea838b7650ab5bb10d089e1ea8613ba2592b05/core/src/main/java/com/datastax/oss/driver/internal/core/os/Native.java#L74)... so I'm inclined to mirror that.

pom.xml
```diff
@@ -36,7 +36,7 @@
     <maven-surefire-plugin.version>3.0.0-M4</maven-surefire-plugin.version>
     <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
-    <datastax-java-driver.version>4.5.0</datastax-java-driver.version>
+    <datastax-java-driver.version>4.6.1</datastax-java-driver.version>
```
tomekl007, line 39: The newest driver didn't introduce any problems in the native build?

absurdfarce: No, with the substitutions in this PR there wasn't a problem at all.  It's as if the native code just always fails to load, so nothing winds up using it (since all calls to it in the current driver code are wrapped by Native.is*Available() checks).

JAVA-2663 will provide the real fix here (via Graal-friendly Native methods)... this just allows us to change driver versions freely without worrying about breaking things.

PR #17, opened 2020-05-20
JAVA-2782: Substitution to fix warning re: Guava's UnsignedBytes impl

absurdfarce:
This PR contains the Guava substitution from https://github.com/datastax/cassandra-quarkus/pull/14

### DIFF
--- a/runtime/src/main/java/com/datastax/oss/driver/shaded/guava/common/primitives/UnsignedBytesSubstitutions.java
+++ b/runtime/src/main/java/com/datastax/oss/driver/shaded/guava/common/primitives/UnsignedBytesSubstitutions.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright DataStax, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.datastax.oss.driver.shaded.guava.common.primitives;
+
+import com.oracle.svm.core.annotate.Alias;
+import com.oracle.svm.core.annotate.Delete;
+import com.oracle.svm.core.annotate.RecomputeFieldValue;
+import com.oracle.svm.core.annotate.Substitute;
+import com.oracle.svm.core.annotate.TargetClass;
+import java.util.Comparator;
+
+@TargetClass(UnsignedBytes.LexicographicalComparatorHolder.UnsafeComparator.class)
+@Delete
+final class UnsafeComparatorSubstitutions {}
+
+@TargetClass(UnsignedBytes.LexicographicalComparatorHolder.class)
+final class LexicographicalComparatorHolderSubstitutions {
+
+  @Alias
+  @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.FromAlias)
+  static Comparator<byte[]> BEST_COMPARATOR = UnsignedBytes.lexicographicalComparatorJavaImpl();
+
+  /* All known cases should be covered by the field substitution above... keeping this only
+   * for sake of completeness */
+  @Substitute
+  static Comparator<byte[]> getBestComparator() {
+    return UnsignedBytes.lexicographicalComparatorJavaImpl();
+  }
+}


adutra:
Actually it just occurred to me @absurdfarce : the fix doesn't belong in the driver either, but rather in [here](https://github.com/datastax/java-driver-shaded-guava). We need to publish a new version of the shaded guava artifact with this substitution in place.

absurdfarce:
@adutra I thought about putting this into the shaded Guava artifact but I had some concerns about that that convinced me it wasn't a good idea.  I definitely agree it's a cleaner solution, however, so convince me these things don't matter and we'll do that. :)

First was the question of versioning.  The shaded Guava artifact is versioned corresponding to the version of Guava it shades, which is entirely reasonable.  But we don't want to just deploy a new version of the 25.1 artifact with these substitutions included since (a) we are actually changing something (and we want users to be able to access the version with and without the substitution) and (b) the artifact isn't _just_ a shaded version of 25.1 anymore.  So we need to change the version in _some_ way.

We can't just change it to 25.1.1 or something; that seems to suggest this corresponds to a _Guava_ version name 25.1.1, which doesn't exist.  So we'd have to use some kind of qualifier, something like 25.1-graal, to clearly indicate what's going on.  And we'd then have to maintain that "-graal" qualifier indefinitely going forward.

My second concern was that this _forces_ the user to leverage this behaviour if they use our shaded Guava artifact.  I liked deploying it at the driver level since it represents the "consumer" of Guava (the driver in this case) making a decision about what behaviour we want.  If we do it at the Guava level _every_ consumer gets this behaviour across the board.  That may or may not be desirable, but they get no choice in the matter.

I'm not sure either of these are fatal necessarily... but there was enough uncertainty there for me to stay away from the idea.

adutra:
> First was the question of versioning.

Thought about that too. In hindsight, I think it is a mistake to version this project after the corresponding Guava version. I wonder if we can switch to a better versioning scheme, e.g. `1.0.0-guava25.1-jre-SNAPSHOT`. @olim7t would you be OK with this change?

> this forces the user to leverage this behaviour if they use our shaded Guava artifact.

Yes, but the description of the artifact is very explicit: 

> Shaded Guava artifact _for use in the DataStax Java driver for Apache Cassandra®_.

So the users are just us :-)

absurdfarce:
I very much like the idea of decoupling version numbers for the shaded JAR from the underlying Guava version... but I'm somewhat skeptical about the specific convention you proposed @adutra.  Specifically I'm thinking specifying the Guava version in this way is redundant.  It seems to suggest that we could have the full range of semver changes (major/minor/patch) within a given version of Guava (i.e. 1.0.0-guava25.1, 1.2.0-guava25.1, 2.3.0-guava25.1).  This seems extraordinarily unlikely to me.

I guess I'd argue for just using conventional semver (major.minor.patch) with a base of 1.0.0 = guava25.1 with no modifications.  Our substitution would be 1.1.0.  Any future Guava upgrade would almost by definition be a major update, but we can document this in a changelog or readme or something similar.

runtime/src/main/java/com/datastax/oss/driver/shaded/guava/common/primitives/UnsignedBytesSubstitutions.java
```diff
@@ -0,0 +1,42 @@
...
+@TargetClass(UnsignedBytes.LexicographicalComparatorHolder.UnsafeComparator.class)
+@Delete
+final class UnsafeComparatorSubstitutions {}
+
+@TargetClass(UnsignedBytes.LexicographicalComparatorHolder.class)
+final class LexicographicalComparatorHolderSubstitutions {
+
+  @Alias
+  @RecomputeFieldValue(kind = RecomputeFieldValue.Kind.FromAlias)
+  static Comparator<byte[]> BEST_COMPARATOR = UnsignedBytes.lexicographicalComparatorJavaImpl();
```
tomekl007, line 34: Which class access was the main problem? Is it `ByteOrder` that uses `jdk.internal.misc.Unsafe`?

adutra:
> It seems to suggest that we could have the full range of semver changes (major/minor/patch) within a given version of Guava (i.e. 1.0.0-guava25.1, 1.2.0-guava25.1, 2.3.0-guava25.1). This seems extraordinarily unlikely to me.

Not really, in Maven's parlance the alphanumeric content that comes after the dash is just a qualifier. I'm using it to clearly state which original version of Guava was used in the process of producing this shaded jar. If you don't put that information in the version number, then you would have to maintain a correspondance table somewhere else, which seems impractical.

runtime/src/main/java/com/datastax/oss/driver/shaded/guava/common/primitives/UnsignedBytesSubstitutions.java
absurdfarce: They have an [impl](https://github.com/google/guava/blob/v25.1/guava/src/com/google/common/primitives/UnsignedBytes.java#L294-L409) of their byte array comparator which uses Unsafe for raw byte operations.  There's also a [pure Java](https://github.com/google/guava/blob/v25.1/guava/src/com/google/common/primitives/UnsignedBytes.java#L411-L430) version available as a fallback, but basically if you're on a 64-bit platform you're pretty likely to get the Unsafe-based version by default.  The [selection logic](https://github.com/google/guava/blob/v25.1/guava/src/com/google/common/primitives/UnsignedBytes.java#L432-L448) relies on a [static init op in the Unsafe impl](https://github.com/google/guava/blob/v25.1/guava/src/com/google/common/primitives/UnsignedBytes.java#L320-L329) to determine whether it's safe to use.  It's the Unsafe.arrayIndexScale() op in that static init op which is the root problem here.  Because that assignment isn't written to a field Graal can't re-write it's value from the build-time architecture (where it's actually computed) to Graal's architecture (which is the value it would need at runtime).  This generates the warnings.

PR #18, opened 2020-05-25
JAVA-2719: Integrate with Netty event loop provided by Quarkus

tomekl007:
This is work in progress because we still didn't decide which path regarding integration with netty we should take.
For more info, see the Mailing List topic: `[quarkus-dev] Cassandra-extension and Quarkus Netty event loops`

todo: 
- [x] take netty event loop from quarkus
- [x] make it opt-out
- [x] use main event loop everywhere (don't use boss)
- [x] switch to newest quarkus release version with all bug fixes related to this ticket
- [x] get rid of `java.util.concurrent.RejectedExecutionException: event executor terminated` warning


olim7t:
I wrote this test resource to check that regular application code doesn't already run on the Netty event loop:
```java
@Path("/threads")
@Produces(MediaType.APPLICATION_JSON)
@Consumes(MediaType.APPLICATION_JSON)
public class ThreadResource {

  @Inject CqlSession session;

  @GET
  public String get() throws ExecutionException, InterruptedException {
    StringBuilder builder =
        new StringBuilder("REST resource runs on: ")
            .append(Thread.currentThread().getName())
            .append('\n');
    CompletionStage<Void> asyncRequest =
        session
            .executeAsync("SELECT release_version FROM system.local")
            .thenApply(
                rs -> {
                  builder.append("Driver I/O runs on: ").append(Thread.currentThread().getName());
                  return null;
                });
    asyncRequest.toCompletableFuture().get();
    return builder.toString();
  }
}
```
Output:
```
REST resource runs on: executor-thread-1
Driver I/O runs on: vert.x-eventloop-thread-11
```
The thread names clearly show that these are two different executors; so we're safe, no risk of deadlocks if the application uses synchronous execute methods from REST resources.

deployment/src/test/java/com/datastax/oss/quarkus/runtime/driver/UseQuarkusNettyEventLoopEnabledTest.java
```diff
@@ -0,0 +1,48 @@
...
+              () -> ShrinkWrap.create(JavaArchive.class).addClasses(CassandraTestBase.class))
+          .withConfigurationResource("application-quarkus-netty-enabled.properties");
+
+  @Test
+  public void should_use_quarkus_netty_event_loop() {
+    // when
+    NettyOptions nettyOptions = ((QuarkusDriverContext) cqlSession.getContext()).getNettyOptions();
+
+    // then
+    assertThat(nettyOptions).isInstanceOf(QuarkusNettyOptions.class);
```
olim7t, line 46: Maybe those test could go a little further: capture a driver I/O thread, using an async request callback, like the example in my previous comment. Then assert that it's from the expected event loop: you could use the thread names, or the fact that default driver threads implement `com.datastax.oss.driver.internal.core.util.concurrent.BlockingOperation.InternalThread`. Or both.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/driver/QuarkusNettyOptions.java
```diff
@@ -0,0 +1,164 @@
...
+  public QuarkusNettyOptions(
+      QuarkusDriverContext context,
+      EventLoopGroup ioEventLoopGroup,
+      EventLoopGroup adminEventLoopGroup) {
+    this.config = context.getConfig().getDefaultProfile();
+    boolean daemon = config.getBoolean(DefaultDriverOption.NETTY_DAEMON);
+    // the NETTY_IO_SHUTDOWN_QUIET_PERIOD, NETTY_IO_SHUTDOWN_TIMEOUT, NETTY_IO_SHUTDOWN_UNIT,
+    // NETTY_ADMIN_SHUTDOWN_QUIET_PERIOD,
+    // NETTY_ADMIN_SHUTDOWN_TIMEOUT, NETTY_ADMIN_SHUTDOWN_UNIT
+    // are ignored when using quarkus netty even loops
```
adutra, line 60:
```suggestion
    // The following options are ignored when using Quarkus Netty event loops:
    // NETTY_IO_SHUTDOWN_QUIET_PERIOD
    // NETTY_IO_SHUTDOWN_TIMEOUT
    // NETTY_IO_SHUTDOWN_UNIT
    // NETTY_ADMIN_SHUTDOWN_QUIET_PERIOD
    // NETTY_ADMIN_SHUTDOWN_TIMEOUT
    // NETTY_ADMIN_SHUTDOWN_UNIT
```

runtime/src/main/java/com/datastax/oss/quarkus/runtime/driver/QuarkusNettyOptions.java
```diff
@@ -0,0 +1,164 @@
...
+    this.ioEventLoopGroup = ioEventLoopGroup;
+    this.adminEventLoopGroup = adminEventLoopGroup;
+
+    // setup the Timer
+    ThreadFactory timerThreadFactory =
+        new ThreadFactoryBuilder()
+            .setThreadFactory(safeFactory)
+            .setNameFormat(context.getSessionName() + "-timer-%d")
+            .setDaemon(daemon)
+            .build();
```
adutra, line 73:
```suggestion
    this.ioEventLoopGroup = ioEventLoopGroup;
    this.adminEventLoopGroup = adminEventLoopGroup;

    // setup the Timer
    ThreadFactory timerThreadFactory =
        new ThreadFactoryBuilder()
            .setThreadFactory(new BlockingOperation.SafeThreadFactory())
            .setNameFormat(context.getSessionName() + "-timer-%d")
            .setDaemon(config.getBoolean(DefaultDriverOption.NETTY_DAEMON))
            .build();
```

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientBuildTimeConfig.java
```diff
@@ -61,4 +61,12 @@
    */
   @ConfigItem(name = "protocol.compression", defaultValue = "none")
   public String protocolCompression;
+
+  /**
+   * Whether or not the DataStax java-driver should use the netty event loop provided by the Quarkus
```
adutra, line 66:
```suggestion
   * Whether or not the DataStax Java driver should use the Netty event loop provided by the Quarkus
```

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientBuildTimeConfig.java
```diff
@@ -61,4 +61,12 @@
...
+   * framework. The default is true, meaning that the java-driver will not construct event loops
```
adutra, line 67:
```suggestion
   * framework. The default is true, meaning that the Java driver will not create event loops
```

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientBuildTimeConfig.java
```diff
@@ -61,4 +61,12 @@
...
+   * itself. When set to true, the {@code QuarkusCqlSession} uses the Quarkus MainEventLoop.
```
adutra, line 68:
```suggestion
   * itself. When set to true, {@code QuarkusCqlSession} will use the main event loop provided by Quarkus.
```

deployment/src/main/java/com/datastax/oss/quarkus/deployment/CassandraClientProcessor.java
```diff
@@ -179,55 +180,20 @@ private void createCassandraClientProducerBean(
         // make CqlSession as Unremovable bean
```
adutra, line 180:
```suggestion
        // mark CqlSession as Unremovable bean
```

deployment/src/test/java/com/datastax/oss/quarkus/runtime/driver/UseQuarkusNettyEventLoopEnabledTest.java
```diff
@@ -0,0 +1,68 @@
...
+    // when
+    AtomicReference<Class<? extends Thread>> threadClass = new AtomicReference<>();
+    CompletionStage<Void> asyncRequest =
+        cqlSession
+            .executeAsync("SELECT release_version FROM system.local")
+            .thenApply(
+                rs -> {
+                  threadClass.set(Thread.currentThread().getClass());
+                  return null;
+                });
```
adutra, line 62:
```suggestion
            .thenAccept(rs -> threadClass.set(Thread.currentThread().getClass()));
```

deployment/src/test/java/com/datastax/oss/quarkus/runtime/driver/UseQuarkusNettyEventLoopDisabledTest.java
```diff
@@ -0,0 +1,69 @@
...
+    // when
+    AtomicReference<Class<? extends Thread>> threadClass = new AtomicReference<>();
+    CompletionStage<Void> asyncRequest =
+        cqlSession
+            .executeAsync("SELECT release_version FROM system.local")
+            .thenApply(
+                rs -> {
+                  threadClass.set(Thread.currentThread().getClass());
+                  return null;
+                });
```
adutra, line 62:
```suggestion
            .thenAccept(rs -> threadClass.set(Thread.currentThread().getClass()));
```

quickstart/pom.xml
```diff
@@ -56,6 +56,7 @@
     <dependency>
       <groupId>io.rest-assured</groupId>
       <artifactId>rest-assured</artifactId>
+      <version>4.2.0</version>
```
adutra, line 59: Remind me: what was the reason why we had to hard-code the version here? We could maybe add a comment explaining this.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/driver/QuarkusNettyOptions.java
```diff
@@ -0,0 +1,164 @@
...
+
+    Duration tickDuration = config.getDuration(DefaultDriverOption.NETTY_TIMER_TICK_DURATION);
+    // JAVA-2264: tick durations on Windows cannot be less than 100 milliseconds,
+    // see https://github.com/netty/netty/issues/356.
+    if (PlatformDependent.isWindows() && tickDuration.toMillis() < 100) {
```
adutra, line 78: A bit sad the we need to repeat this code here.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/driver/QuarkusNettyOptions.java
tomekl007: This is copy-paste from the original `DefaultNettyOptions` so I would like to keep it in sync if possible

quickstart/pom.xml
tomekl007: It is not provided in the quarkus bom. I can extract it to a parameter if you want

runtime/src/main/java/com/datastax/oss/quarkus/runtime/driver/QuarkusNettyOptions.java
tomekl007: Agree, there is some copy-paste here form `DefaultNettyOptions`.
We could refactor the original `DefaultNettyOptions` to:
```
  // setup the Timer
    ThreadFactory timerThreadFactory =
        new ThreadFactoryBuilder()
            .setThreadFactory(safeFactory)
            .setNameFormat(context.getSessionName() + "-timer-%d")
            .setDaemon(daemon)
            .build();

    this.timer = constructTimer(timerThreadFactory);
  }

  protected Timer constructTimer(ThreadFactory timerThreadFactory)
  {
    Duration tickDuration = config.getDuration(DefaultDriverOption.NETTY_TIMER_TICK_DURATION);
    // JAVA-2264: tick durations on Windows cannot be less than 100 milliseconds,
    // see https://github.com/netty/netty/issues/356.
    if (PlatformDependent.isWindows() && tickDuration.toMillis() < 100) {
      LOG.warn(
          "Timer tick duration was set to a value too aggressive for Windows: {} ms; "
              + "doing so is known to cause extreme CPU usage. "
              + "Please set advanced.netty.timer.tick-duration to 100 ms or higher.",
          tickDuration.toMillis());
    }
    return 
        new HashedWheelTimer(
            timerThreadFactory,
            tickDuration.toNanos(),
            TimeUnit.NANOSECONDS,
            config.getInt(DefaultDriverOption.NETTY_TIMER_TICKS_PER_WHEEL));
  }
```
and use the `constructTimer()` method here. wdyt?

adutra: Well you said you prefer to keep the copy-paste in sync so 🤷‍♂️ 
In any case, inheriting from `DefaultNettyOptions` is tricky because all the stuff that happens in its constructor. A better solution imo would be to isolate the Timer creation into a factory class.

quickstart/pom.xml
adutra: Hmm I remember there was something else... @absurdfarce do you remember? But for now let's keep it the way it is, it's just a detail.

PR #61, opened 2020-06-02
Validate that cassandra-quarkus/runtime module does not have a dependency to quarkus-core-deployment

tomekl007:


runtime/pom.xml
```diff
@@ -91,9 +91,12 @@
       <optional>true</optional>
     </dependency>
     <dependency>
-      <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-smallrye-metrics-spi</artifactId>
-      <version>${quarkus.version}</version>
+      <groupId>org.eclipse.microprofile.metrics</groupId>
+      <artifactId>microprofile-metrics-api</artifactId>
+    </dependency>
+    <dependency>
```
adutra, line 97: 1. What is the purpose of this dependency? I tried removing it and didn't notice any difference.
2. In any case, it should probably be in `<scope>runtime</scope>`.

tomekl007: We need it in the compile scope for the native mode

adutra: Can you elaborate? I successfully executed all the native tests without this dependency.

tomekl007: Are you sure?
I am running `mvn clean verify -Dnative` and getting:
```
java.lang.ClassNotFoundException: org.eclipse.microprofile.metrics.MetricRegistry
```
without this dependency. It was also the same situation on jenkins

adutra: I don't see this, however I found an issue indeed, but it's with the quickstart module.

I think we should remove this dependency from here since the runtime module shouldn't impose any logging framework on users.

However I suggest that we add this to quickstart pom:

```
    <!-- Tinkerpop uses JCL directly; redirect JCL calls to JBoss logging  -->
    <dependency>
      <groupId>org.jboss.logging</groupId>
      <artifactId>commons-logging-jboss-logging</artifactId>
      <scope>runtime</scope>
    </dependency>
```

adutra: Hmm thinking about this a bit more, maybe it's better to impose the usage of the JCL-to-JBoss bridge provided by this dependency, since Tinkerpop will need it anyways. But I still think that this should be in runtime scope.

tomekl007: ok, done

PR #64, opened 2020-06-08
JAVA-2778: Split classes into api and internal packages

tomekl007:


deployment/src/test/java/com/datastax/oss/quarkus/runtime/internal/health/CassandraHealthBuildItemEnabledTest.java
```diff
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.datastax.oss.quarkus.runtime.health;
+package com.datastax.oss.quarkus.runtime.internal.health;
```
adutra, line 16: I still don't see why we need to keep package names containing `runtime` for tests in the `deployment` module. I just tried moving this class and a few others and everything works just fine:

More specifically I placed the following classes in `com.datastax.oss.quarkus.deployment.internal`: 

1. CassandraHealthBuildItemEnabledTest
1. CassandraHealthCheckBuildItemDisabledTest
1. CassandraMetricsTest
1. QuarkusCqlSessionTest
1. UseQuarkusNettyEventLoopDisabledTest
1. UseQuarkusNettyEventLoopEnabledTest

tomekl007: Those are tests that are examining the components from the `runtime` so I think it is better to have the runtime package name there. I saw a similar pattern in some extensions.

I agree to create ta `demo` package, for `tests` i am not sure because we have the "production code" and tests code. Having the package named `tests` for both may be misleading.

adutra: Please, no split packages. I would really prefer that 1 module = 1 _distinct_ root package.

Granted, we are probably never (never?) making this project OSGi compatible, but people could reasonably ask, I think, that this project be made compatible with the Java module system. Because of that, I think that we should not mix packages now, we could regret it later.

If some extensions are mixing packages, that's their problem, it's not imo a pattern that we absolutely need to follow.

adutra:
I think it's better to merge this one sooner than later as it will create conflicts with other PRs.

PR #66, opened 2020-06-09
JAVA-2761  Improve in-tree documentation 

tomekl007:

PR #67, opened 2020-06-09
JAVA-2791: Add IT test to the quickstart project

tomekl007:


integration-tests/pom.xml
```diff
@@ -122,6 +122,17 @@
   </dependencies>
   <build>
     <plugins>
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
```
adutra, line 126: This won't work very well because artifacts for this module are not meant to be deployed to Maven Central, see [here]( https://github.com/datastax/java-driver/blob/2997d775a091cfee614e49d20c98f5b29518d7c0/integration-tests/pom.xml#L318-L323).

As far as i can tell, you only need one class in this module: `CassandraTestResource`. I'm afraid you are going to have to copy-paste it.

It turns out that we already have 2 copies of this class:

1. deployment/src/test/java/com/datastax/oss/quarkus/CassandraTestBase.java
2. integration-tests/src/test/java/com/datastax/oss/quarkus/CassandraTestResource.java

So I suggest that you create a third copy for now. But we definitely should look into creating a `test-infra` module to host test utilities like this one.

integration-tests/src/main/resources/init_script.cql
```diff
@@ -0,0 +1,3 @@
+CREATE KEYSPACE IF NOT EXISTS k1 WITH replication = {'class':'SimpleStrategy', 'replication_factor':1};
+CREATE TABLE IF NOT EXISTS k1.test_NameConverterEntity(test_entityId int primary key);
```
adutra, line 2: Not related to this PR but just because I stumble upon it: very weird table name (used mixed-case letters, not recommended) and very weird schema (only one partition key column).

quickstart/pom.xml
```diff
@@ -87,6 +87,18 @@
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-resteasy-mutiny</artifactId>
     </dependency>
+    <dependency>
```
adutra, line 90: As mentioned above, probably better to get rid of this.

quickstart/src/test/java/com/datastax/oss/quarkus/FruitResourceIT.java
```diff
@@ -0,0 +1,65 @@
...
+import org.json.JSONException;
+import org.json.JSONObject;
+import org.junit.jupiter.api.Test;
+
+@QuarkusTest
+@QuarkusTestResource(CassandraTestResource.class)
+public class FruitResourceIT {
+
+  @Test
+  public void testSaveAndRetrieveProduct() throws JSONException {
```
adutra, line 36: These tests are not following the naming conventions that you are adding in #66.

integration-tests/src/test/java/com/datastax/oss/quarkus/CassandraTestResource.java
```diff
@@ -20,16 +20,16 @@
...
 
 public class CassandraTestResource implements QuarkusTestResourceLifecycleManager {
   private static final Logger LOGGER = Logger.getLogger(CassandraTestResource.class);
-  private static GenericContainer<?> cassandraContainer;
+  private static CassandraContainer<?> cassandraContainer;
 
   @Override
   public Map<String, String> start() {
     cassandraContainer = new CassandraContainer<>();
+    cassandraContainer.withInitScript("init_script.cql");
```
adutra, line 32: Nice 👍

integration-tests/pom.xml
tomekl007: It does work, see:
https://jenkins-drivers.build.dsinternal.org/job/tools/job/cassandra-quarkus/job/java2791/3/testReport/com.datastax.oss.quarkus/
The FruitIT is executed succesfully

integration-tests/src/main/resources/init_script.cql
tomekl007: This name is done on purpose to exercise the `customConverterClass` and `TestNameConverter` that uses `@RegisterForReflection` with mapper

integration-tests/pom.xml
adutra: But it won't if someone copies the quickstart and uses it as a standalone project. The test-jar is coming from the build itself, not from Maven Central. So the quickstart cannot be a standalone project anymore, if you keep this dependency.

integration-tests/src/main/resources/init_script.cql
adutra: ok, not a problem for this PR anyways.

integration-tests/pom.xml
tomekl007: ah. you are right. Done

PR #73, opened 2020-06-15
add configuration for username and password

thomasbecker:
Basically I needed that myself and looked for existing issues corresponding to this. Found 2697 and here's a pull request for at least username and password.


runtime/src/main/java/com/datastax/oss/quarkus/runtime/api/config/CassandraClientConnectionConfig.java
```diff
@@ -39,4 +39,15 @@
   /** How long the driver waits for a request to complete. */
   @ConfigItem(name = "request.timeout")
   public Optional<Duration> requestTimeout;
+
+  /** The auth provider class. If not specified, PlainTextAuthProvider will be used */
+  public Optional<String> authProviderClass;
```
adutra, line 44: I think the auth provider class is not useful here, because the only settings exposed are `username` and `password`, which means that the only auth provider that can work is `PlainTextAuthProvider` anyways. I suggest that instead you set this option to `PlainTextAuthProvider` in `CassandraClientProducer` if `username` and `password` were provided.

thomasbecker: I think it defaults to PlainTextAuthProvider. Let me verify that and then I will simply drop the auth provider setting.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientProducer.java
```diff
@@ -149,6 +149,12 @@ private void configureConnectionSettings(
         DefaultDriverOption.LOAD_BALANCING_LOCAL_DATACENTER, config.localDatacenter);
     config.requestTimeout.ifPresent(
         v -> configLoaderBuilder.withDuration(DefaultDriverOption.REQUEST_TIMEOUT, v));
+    config.username.ifPresent(
+        username ->
+            configLoaderBuilder.withString(DefaultDriverOption.AUTH_PROVIDER_USER_NAME, username));
+    config.password.ifPresent(
+        password ->
+            configLoaderBuilder.withString(DefaultDriverOption.AUTH_PROVIDER_PASSWORD, password));
```
adutra, line 157: If you don't specify the AuthProvider class here, the driver won't create any for you.

```suggestion
    if (config.username.isPresent() && config.password.isPresent()) {
      configLoaderBuilder
          .withClass(DefaultDriverOption.AUTH_PROVIDER_CLASS, PlainTextAuthProvider.class)
          .withString(DefaultDriverOption.AUTH_PROVIDER_USER_NAME, config.username.get())
          .withString(DefaultDriverOption.AUTH_PROVIDER_PASSWORD, config.password.get());
    }
```

adutra:
@thomasbecker would it be possible for you to add a small test in `com.datastax.oss.quarkus.deployment.internal.CassandraClientConfigTest`? Otherwise don't worry, we can do that.

adutra:
Basically add:

`application-cassandra-client.properties`:
```
quarkus.cassandra.username=alice
quarkus.cassandra.password=fakePasswordForTests
```

`CassandraClientConfigTest`:
```
  @Test
  public void should_enable_auth() {
    DriverExecutionProfile profile = cqlSession.getContext().getConfig().getDefaultProfile();
    assertThat(profile.getString(AUTH_PROVIDER_CLASS))
        .isEqualTo(PlainTextAuthProvider.class.getName());
    assertThat(profile.getString(AUTH_PROVIDER_USER_NAME)).isEqualTo("alice");
    assertThat(profile.getString(AUTH_PROVIDER_PASSWORD)).isEqualTo("fakePasswordForTests");
  }
```

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
```diff
@@ -100,7 +102,9 @@
         // timestamp generators
         new ReflectiveClassBuildItem(true, true, AtomicTimestampGenerator.class.getName()),
         new ReflectiveClassBuildItem(true, true, ThreadLocalTimestampGenerator.class.getName()),
-        new ReflectiveClassBuildItem(true, true, Publisher.class.getName()));
+        new ReflectiveClassBuildItem(true, true, Publisher.class.getName()),
+        // authentication
+        new ReflectiveClassBuildItem(true, true, PlainTextAuthProvider.class.getName()));
```
adutra, line 107: Good catch! It's a pitty that we cannot (yet) use TestContainers with authentication. Let's add all the default auth providers that ship with the driver.

```suggestion
        // reactive streams
        new ReflectiveClassBuildItem(true, true, Publisher.class.getName()),
        // authentication
        new ReflectiveClassBuildItem(true, true, PlainTextAuthProvider.class.getName()));
        new ReflectiveClassBuildItem(true, true, DseGssApiAuthProvider.class.getName()));
        new ReflectiveClassBuildItem(true, true, ProgrammaticPlainTextAuthProvider.class.getName()));
        new ReflectiveClassBuildItem(true, true, ProgrammaticDseGssApiAuthProvider.class.getName()));
```

pom.xml
```diff
@@ -26,6 +26,18 @@
   <description>Cassandra extensions for Quarkus</description>
   <url>https://quarkus.io/guides/cassandra</url>
   <inceptionYear>2020</inceptionYear>
+  <distributionManagement>
```
adutra, line 29: Not sure what this is for?

thomasbecker: Not supposed to go here. 👍 Will be removed in a second.

thomasbecker:
I'm still in the process of testing this end to end. I hope I can finish it by today and will report back once I'm happy.

tomekl007:
> I'm still in the process of testing this end to end. I hope I can finish it by today and will report back once I'm happy.

Please run `mvn fmt:format` to format your code according to our contribution guidelines.

thomasbecker:
@tomekl007 Yeah, I know. My fault. I accepted the additions from Alexandre without reviewing and reformatting. Sloppy on my side. Still didn't have a chance to test it end 2 end in a native image. 

Seems like the native image starts so fast, that network is not yet properly available on the pod in gke (never seen that before). Basically the java driver complains about unknown host. When I ssh into the machine the host is resolved just fine.  So connecting to cassandra fails. Working on it. Will keep you guys updated.

thomasbecker:
Finally I had some time to test it end to end on my local box. Plaintext auth works fine in a "normal" JDK. Also works fine in a native image locally and on some production like gke cluster.

When debug logs are enabled I found some more classes that needed to be registered for reflection. However all my use cases (pretty basic, not much more complicated than what the quickstart app does) do work fine with or without those registrations. The ClassNotFound Errors just appear in the debug logs, so not sure if this is an issue for other use cases.

Fixed them by adding: 
        new ReflectiveClassBuildItem(true, true, OGCGeometry.class.getName()),
        new ReflectiveClassBuildItem(true, true, TinkerIoRegistryV3d0.class.getName()),
        new ReflectiveClassBuildItem(true, true, JsonParser.class.getName()),
        new ReflectiveClassBuildItem(true, true, GraphTraversal.class.getName()));

Let me know if you want those registrations in the code. I will create another pull request to separate them from the auth stuff if you do want them.

So this one here is working fine from my perspective. Feel free to merge it or let me know if we need to add something. Thx for the support and good work on the extension, guys!

adutra:
Well thank YOU for being available and for testing our extension even before the first alpha release 👍 !

I suggest that we focus back on the initial purpose of this PR, which is to add support for plain text authentication. @tomekl007 is already trying to solve a few missing registrations in #78 , so I would suggest that we tackle registrations there, not here.

That said, I think I know where these missing registrations come from: it's in `com.datastax.oss.driver.internal.core.util.DependencyCheck` – I will let @tomekl007 add them in his PR.

adutra:
@thomasbecker I polished a bit your PR and committed as 54620be84a9c751a3197e679a1102dc36b5e84fa. Thank you!

thomasbecker:
My pleasure. Thank you.

PR #74, opened 2020-06-15
Upgrade to Java driver 4.7.0 as released

absurdfarce:


runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientProducer.java
```diff
@@ -93,9 +93,9 @@ private ProgrammaticDriverConfigLoaderBuilder createDriverConfigLoader() {
               .withFallback(ConfigFactory.parseResources("application.conf"))
               .withFallback(ConfigFactory.parseResources("application.json"))
               .withFallback(ConfigFactory.defaultReference())
+              .getConfig(DefaultDriverConfigLoader.DEFAULT_ROOT_PATH)
               .resolve();
-        },
-        DefaultDriverConfigLoader.DEFAULT_ROOT_PATH) {
+        }) {
```
absurdfarce, line 98: Change above required since 4.7.0 does away with DefaultProgrammaticDriverConfigLoaderBuilder(@NonNull Supplier<Config> fallbackSupplier, String root) in favor of https://github.com/datastax/java-driver/blob/4.7.0/core/src/main/java/com/datastax/oss/driver/internal/core/config/typesafe/DefaultProgrammaticDriverConfigLoaderBuilder.java#L82-L84

PR #76, opened 2020-06-17
issue59 - Create automated performance tests for the extension 

tomekl007:


performance/falloutPerformanceTest.yaml
```diff
@@ -0,0 +1,183 @@
...
+        cloud.instance.type: ms1.small
+        name: dseserver-cassandra-quarkus-perf
+        mark_for_reuse: true
+    configuration_manager:
+      - name: ctool
+        properties:
+          java.version: "1.8_151"
+          product.type: dse
+          product.install.type: tarball
+          product.version: 6.7-dev
```
adutra, line 30: I wonder if OSS Cassandra isn't enough?

performance/falloutPerformanceTest.yaml
```diff
@@ -0,0 +1,183 @@
...
+        module: bash
+        properties:
+          target.group: quickstartapp
+          export_output: false
+          timeout: 1 hours
+          script: |
+            cd ${FALLOUT_SCRATCH_DIR}
+            sudo apt update --assume-yes
+            sudo apt-get install lftp --assume-yes
+            sudo apt-get install zlib1g-dev libxmu-dev libxmu-headers freeglut3-dev libxext-dev libxi-dev --assume-yes
```
adutra, line 117: Wow, why do we need all of this?

performance/falloutPerformanceTest.yaml
```diff
@@ -0,0 +1,183 @@
...
+        module: bash
+        properties:
+          target.group: http-client
+          export_output: false
+          timeout: 1 hours
+          script: |
+            cd ${FALLOUT_SCRATCH_DIR}
+            sudo apt update --assume-yes
+            sudo apt-get install lftp --assume-yes
+            sudo apt-get install wrk --assume-yes
```
adutra, line 165: Glad to see that it exists in apt!

performance/falloutPerformanceTest.yaml
```diff
@@ -0,0 +1,183 @@
+# branch and version of the cassandra-extension to benchmark
```
adutra, line 1: `fallout.yaml` for the file name is enough imo.

performance/falloutPerformanceTest.yaml
tomekl007: those are needed for native graal compilation

performance/falloutPerformanceTest.yaml
tomekl007: If we would like to test graph or continuous paging in the future maybe it is better to use DSE

PR #77, opened 2020-06-18
issue70 - Clarify usage of log frameworks

tomekl007:


CONTRIBUTING.adoc
```diff
@@ -103,11 +103,12 @@ explains where the component fits in the architecture, and anything else that yo
 
 === Logs
 
-We use SLF4J; loggers are declared like this:
+We use jboss logger; loggers are declared like this:
```
adutra, line 106: So i'm hesitating. According to [this](https://quarkus.io/guides/logging) and [this](https://antoniogoncalves.org/2019/06/07/configuring-a-quarkus-application/) we have many choices, and SLF4J is supported.

tomekl007: When analyzing other extensions in the quarkus repository you will notice that all of them are using JBoss instead of slf4j

adutra: But wouldn't it be safer to just keep the current code? 

The only thing I was planning to add is dependency declarations in all modules using SLF4J:

```
      <dependency>
        <groupId>org.slf4j</groupId>
        <artifactId>slf4j-api</artifactId>
      </dependency>
      <dependency>
        <groupId>org.jboss.logging</groupId>
        <artifactId>jboss-logging</artifactId>
        <scope>runtime</scope>
      </dependency>
```

tomekl007: Other extensions are using jboss logger API (i.e `debugf` instead of `debug`)

tomekl007: - sl4j is already excluded from java-driver in the quarkus - I've excluded the slf4j form java-driver in the `quickstart`
- I am not sure if we need to add the dependency to JBoss explicitly. It is already in the quarkus dependency
wdyt?

CONTRIBUTING.adoc
```diff
@@ -103,11 +103,12 @@ explains where the component fits in the architecture, and anything else that yo
...
 
 [source,java]
 ----
-private static final Logger LOG = LoggerFactory.getLogger(TheEnclosingClass.class);
+import org.jboss.logging.Logger;
+private static final Logger LOG = Logger.getLogger(TheEnclosingClass.class.getName());
```
adutra, line 111:
```suggestion
private static final Logger LOG = Logger.getLogger(TheEnclosingClass.class);
```

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/driver/QuarkusNettyOptions.java
```diff
@@ -38,11 +38,10 @@
 import java.time.Duration;
 import java.util.concurrent.ThreadFactory;
 import java.util.concurrent.TimeUnit;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.jboss.logging.Logger;
 
 public class QuarkusNettyOptions implements NettyOptions {
-  private static final Logger LOG = LoggerFactory.getLogger(QuarkusNettyOptions.class);
+  private static final Logger LOG = Logger.getLogger(QuarkusNettyOptions.class.getName());
```
adutra, line 44:
```suggestion
  private static final Logger LOG = Logger.getLogger(QuarkusNettyOptions.class);
```

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/metrics/MicroProfileSessionMetricUpdater.java
```diff
@@ -33,12 +32,12 @@
...
 import org.eclipse.microprofile.metrics.MetricRegistry;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
+import org.jboss.logging.Logger;
 
 public class MicroProfileSessionMetricUpdater extends MicroProfileMetricsUpdater<SessionMetric>
     implements SessionMetricUpdater {
-  private static final Logger LOG = LoggerFactory.getLogger(DropwizardSessionMetricUpdater.class);
+  private static final Logger LOG =
+      Logger.getLogger(MicroProfileSessionMetricUpdater.class.getName());
```
adutra, line 40:
```suggestion
      Logger.getLogger(MicroProfileSessionMetricUpdater.class);
```

CONTRIBUTING.adoc
adutra: > it is usually considered good manners to always declare dependencies that are directly used by the module.

Although I confess that the Quarkus framework doesn't seem to follow this pattern, so I'm OK with not adding the dependency explciitly.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/driver/QuarkusNettyOptions.java
tomekl007: Why?
the `getLogger` is a proxy method that does what we need:
```
public static Logger getLogger(Class<?> clazz) {
        return getLogger(clazz.getName());
}
```

adutra: Because it's less code?

adutra:
I'm sorry but I think I'm changing my mind here. The more I analyze how logging is supposed to work the more I think we should stick with SLF4J because:

1. It's what the driver uses anyways.
2. Using JBoss Logging is only a requirement if you inherit from Quarkus parent pom, which is not our case. When inheriting from Quarkus parent, it makes sense to exclude SLF4J since other logging frameworks (JBoss) are already present. But in our case no logging framework is present by default, so we really can choose whichever we want IMO.
3. Regarding JCL bridge over SLF4J, it also makes sense to exclude it when inheriting from Quarkus parent: an equivalent bridge is already present. But in our case it shouldn't be removed because no other bridge is present by default.

adutra:
Also to note: quarkus-core has a dependency on slf4j-api, so I guess that's what they want extensions to use primarily. See also https://github.com/quarkusio/quarkus/issues/294#issuecomment-447591945.

tomekl007:
> Also to note: quarkus-core has a dependency on slf4j-api, so I guess that's what they want extensions to use primarily. See also [quarkusio/quarkus#294 (comment)](https://github.com/quarkusio/quarkus/issues/294#issuecomment-447591945).

I am bit confused because, on the other hand, all extensions in the quarkus repo are using JBoss

adutra:
> I am bit confused because, on the other hand, all extensions in the quarkus repo are using JBoss

From what I see if your extension is in quarkus repo, you have 2 choices: 

1. use sl4j-api as frontend and org.jboss.slf4j:slf4j-jboss-logging as backend (slf4j-api is not banned)
2. use jboss-logging as your frontend and backend.

I guess that out of habitude, everybody went for option 2.

Also, apparently option 1 gets substituted at runtime with option 2. see https://github.com/quarkusio/quarkus/blob/987a4db26d31057ea562d287d9b1d4566bc4a176/core/runtime/src/main/java/io/quarkus/runtime/graal/LoggingSubstitutions.java.

tomekl007:
> Also, apparently option 1 gets substituted at runtime with option 2. see https://github.com/quarkusio/quarkus/blob/987a4db26d31057ea562d287d9b1d4566bc4a176/core/runtime/src/main/java/io/quarkus/runtime/graal/LoggingSubstitutions.java.

yes, I saw that substitution - so it seems that using JBoss is more "friendly" for native graal mode; it does not involve a substitution code

adutra:
@tomekl007 I think this is obsolete now, we verified that using SLF4J is 100% supported. OK to close?

tomekl007:
> @tomekl007 I think this is obsolete now, we verified that using SLF4J is 100% supported. OK to close?

yes - closed

PR #78, opened 2020-06-18
Register for reflection for AuthProviders and SSLFactories

tomekl007:


deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
```diff
@@ -123,6 +130,32 @@
...
+        new ReflectiveClassBuildItem(true, true, PlainTextAuthProvider.class.getName()),
+        new ReflectiveClassBuildItem(true, true, DseGssApiAuthProvider.class.getName()),
+        new ReflectiveClassBuildItem(true, true, ProgrammaticPlainTextAuthProvider.class.getName()),
+        new ReflectiveClassBuildItem(
+            true, true, ProgrammaticDseGssApiAuthProvider.class.getName()));
+  }
+
+  @BuildStep
+  List<ReflectiveClassBuildItem> registerTimestampGeneratorsForReflection() {
+    return Arrays.asList( // timestamp generators
```
adutra, line 153:
```suggestion
    return Arrays.asList(
```

adutra:
As @thomasbecker pointed out in #73, there are a few reflective calls in `com.datastax.oss.driver.internal.core.util.DependencyCheck`:

* "com.esri.core.geometry.ogc.OGCGeometry"
* "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal"
* "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerIoRegistryV3d0"
* "com.fasterxml.jackson.core.JsonParser",
* "com.fasterxml.jackson.databind.ObjectMapper"
* "net.jpountz.lz4.LZ4Compressor"
* "org.xerial.snappy.Snappy" (<-- maybe not this one?)

@tomekl007 wdyt? I think we need to include them as well.

tomekl007:
> @tomekl007 wdyt? I think we need to include them as well.

- The lz4 is already handled by the `registerLz4ForReflection()` method.
- `org.xerial.snappy.Snappy` doesn't need to be included because it is substituted and not supported in the native mode.

Regarding the rest of the classes - I agree. Done in https://github.com/datastax/cassandra-quarkus/pull/78/commits/436d25b1f30db8028e578822813e5b247a154ff0

Issue #82, opened 2020-06-18
Make mapper-runtime dependency optional


adutra:
Currently in the runtime module: java-driver-mapper-runtime is a required dependency. Not every application will use the mapper, it would then make sense to make this dependency optional.

adutra:
Note that query builder is a dependency of mapper-runtime anyways.

Issue #79, opened 2020-06-18
Double-check driver dependency exclusions


adutra:
We currently exclude a lot of driver dependencies:

```
    <dependency>
      <groupId>com.datastax.oss</groupId>
      <artifactId>java-driver-core</artifactId>
      <exclusions>
        <exclusion>
          <groupId>org.slf4j</groupId>
          <artifactId>jcl-over-slf4j</artifactId>
        </exclusion>
        <exclusion>
          <groupId>org.codehaus.groovy</groupId>
          <artifactId>groovy</artifactId>
        </exclusion>
        <exclusion>
          <groupId>org.codehaus.groovy</groupId>
          <artifactId>groovy-json</artifactId>
        </exclusion>
        <exclusion>
          <groupId>com.squareup</groupId>
          <artifactId>javapoet</artifactId>
        </exclusion>
        <exclusion>
          <groupId>org.ow2.asm</groupId>
          <artifactId>asm-util</artifactId>
        </exclusion>
        <exclusion>
          <groupId>io.dropwizard.metrics</groupId>
          <artifactId>metrics-core</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
```

I'm not sure all of this is required. The following seems more appropriate, and the build passes including native tests:

```
    <dependency>
      <groupId>com.datastax.oss</groupId>
      <artifactId>java-driver-core</artifactId>
      <exclusions>
        <exclusion>
          <groupId>io.dropwizard.metrics</groupId>
          <artifactId>metrics-core</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
```

It would be good to investigate what needs to be excluded, and why.

tomekl007:
The reason why they are excluded is that they are part of the `bannedDependencies` in the quarkus:
`https://github.com/quarkusio/quarkus/blob/master/build-parent/pom.xml#L279`

If we will remove those dependencies, our build will pass but we may have a problem with CI build against the quarkus itself (this https://github.com/datastax/cassandra-quarkus/actions?query=workflow%3A%22Quarkus+ecosystem+CI%22)

adutra:
Can we know for sure ? the pom file you linked in your comment is not a parent of our extension.

tomekl007:
I think that we would need to run the build against https://github.com/datastax/cassandra-quarkus/actions?query=workflow%3A%22Quarkus+ecosystem+CI%22.

[...]

PR #89, opened 2020-06-18
Cleanup POM files

adutra:
This commit cleans up POM files, reorders dependencies,
Fixes a few scope problems, revisits driver dependency
exclusions, revisits dependency management and plugin
management sections, and also makes minor cosmetic
improvements to POM files.

### DIFF
--- a/bom-runtime/pom.xml
+++ b/bom-runtime/pom.xml
@@ -37,6 +37,22 @@
         <scope>import</scope>
       </dependency>
       <!-- DataStax Java Driver BOM -->
+      <!--
+      Note: we explicitly declare java-driver-core here,
+      and before the driver bom, in order for
+      the exclusions to be taken into account
+      -->
+      <dependency>
+        <groupId>com.datastax.oss</groupId>
+        <artifactId>java-driver-core</artifactId>
+        <version>${datastax-java-driver.version}</version>
+        <exclusions>
+          <exclusion>
+            <groupId>io.dropwizard.metrics</groupId>
+            <artifactId>metrics-core</artifactId>
+          </exclusion>
+        </exclusions>
+      </dependency>
       <dependency>
         <groupId>com.datastax.oss</groupId>
         <artifactId>java-driver-bom</artifactId>
@@ -44,10 +60,15 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
-      <!-- Cassandra Quarkus Extension (Runtime) -->
+      <!-- Cassandra Quarkus Extension (Runtime & Test Framework) -->
       <dependency>
         <groupId>com.datastax.oss.quarkus</groupId>
         <artifactId>cassandra-quarkus-client</artifactId>
+        <version>${project.version}</version>
+      </dependency>
+      <dependency>
+        <groupId>com.datastax.oss.quarkus</groupId>
+        <artifactId>cassandra-quarkus-test-framework</artifactId>
         <version>${project.version}</version>
       </dependency>
       <!-- Driver Optional dependencies -->
--- a/deployment/pom.xml
+++ b/deployment/pom.xml
@@ -50,19 +50,12 @@
       <artifactId>cassandra-quarkus-client</artifactId>
     </dependency>
     <dependency>
-      <groupId>com.datastax.oss.quarkus</groupId>
-      <artifactId>cassandra-quarkus-test-framework</artifactId>
-      <version>${project.version}</version>
-      <scope>test</scope>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-netty-deployment</artifactId>
     </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-smallrye-health-spi</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-smallrye-health</artifactId>
-      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
@@ -85,11 +78,17 @@
     </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-netty-deployment</artifactId>
+      <artifactId>quarkus-smallrye-health</artifactId>
+      <scope>test</scope>
     </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-smallrye-metrics</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>com.datastax.oss.quarkus</groupId>
+      <artifactId>cassandra-quarkus-test-framework</artifactId>
       <scope>test</scope>
     </dependency>
   </dependencies>
--- a/integration-tests/pom.xml
+++ b/integration-tests/pom.xml
@@ -39,11 +39,11 @@
   <dependencies>
     <dependency>
       <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-resteasy</artifactId>
+      <artifactId>quarkus-resteasy-mutiny</artifactId>
     </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-resteasy-mutiny</artifactId>
+      <artifactId>quarkus-smallrye-health</artifactId>
     </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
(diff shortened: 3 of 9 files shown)


bom-runtime/pom.xml
```diff
@@ -37,6 +37,22 @@
...
+      and before the driver bom, in order for
+      the exclusions to be taken into account
+      -->
+      <dependency>
+        <groupId>com.datastax.oss</groupId>
+        <artifactId>java-driver-core</artifactId>
+        <version>${datastax-java-driver.version}</version>
+        <exclusions>
+          <exclusion>
+            <groupId>io.dropwizard.metrics</groupId>
```
adutra, line 51: I experimented with existing exclusions and as far as I can tell:

1. Excluding jcl-over-slf4j is IMO a bad idea. This is a bridge from JCL to SLF4J, i.e., it provides an implementation of the JCL API that delegates to SLF4J. Excluding this results in `NoClassDefFoundError`, and the only way to fix it is to include another bridge, namely, commons-logging-jboss-logging. But we are just swapping one bridge for another.
2. Other exclusions: groovy, groovy-json, javapoet, asm-util. AFAICT there is no difference. I suggest that we don't exclude them, unless I'm missing something, but in this case it would be nice to document why we exclude them.

deployment/pom.xml
```diff
@@ -79,36 +78,19 @@
     </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-netty-deployment</artifactId>
-    </dependency>
-    <dependency>
-      <groupId>org.testcontainers</groupId>
-      <artifactId>cassandra</artifactId>
-      <version>1.12.4</version>
-      <exclusions>
```
adutra, line 88: Exclusions not required IMO, the tests run fine without them.

integration-tests/pom.xml
```diff
@@ -56,57 +56,26 @@
...
-          <groupId>io.dropwizard.metrics</groupId>
-          <artifactId>metrics-core</artifactId>
-        </exclusion>
-      </exclusions>
     </dependency>
     <dependency>
       <groupId>com.datastax.oss</groupId>
       <artifactId>java-driver-mapper-runtime</artifactId>
     </dependency>
-    <dependency>
```
adutra, line 70: Not required in compile scope.

integration-tests/pom.xml
```diff
@@ -56,57 +56,26 @@
...
-    </dependency>
     <dependency>
       <groupId>org.lz4</groupId>
       <artifactId>lz4-java</artifactId>
+      <scope>runtime</scope>
     </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
       <artifactId>quarkus-test-common</artifactId>
+      <scope>test</scope>
```
adutra, line 72: Was in wrong scope (compile).

integration-tests/pom.xml
```diff
@@ -135,11 +103,6 @@
               <artifactId>java-driver-mapper-processor</artifactId>
               <version>${datastax-java-driver.version}</version>
             </path>
-            <path>
```
adutra, line 138: This won't be required after JAVA-2800, and we can live without it for now (generates a warning).

pom.xml
```diff
@@ -49,17 +53,44 @@
     <module>quickstart</module>
     <module>documentation</module>
   </modules>
+  <dependencyManagement>
+    <dependencies>
```
adutra, line 57: This is IMO the best place to declare test dependencies that we don't want to clutter the BOMs with.

pom.xml
```diff
@@ -49,17 +53,44 @@
...
+        <groupId>org.testcontainers</groupId>
+        <artifactId>cassandra</artifactId>
+        <version>1.12.4</version>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
   <build>
     <pluginManagement>
       <plugins>
+        <plugin>
```
adutra, line 75: Trying to follow the convention of declaring versions inside management sections, but leaving configuration details to actual dependency/plugin declarations.

quickstart/pom.xml
```diff
@@ -17,13 +17,10 @@
 
 -->
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
-  <parent>
```
adutra, line 20: Removing the parent makes this module truly independent, while still building it when the parent module is built. IOW, the dependency parent -> quickstart still exists, but the dependency quickstart -> parent is gone. With this, we can be 100% sure that the project can be used as a standalone project stub. The drawback, but we know this already, is that all plugin and dependency versions must be re-declared here.

quickstart/pom.xml
```diff
@@ -41,90 +38,110 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>com.datastax.oss.quarkus</groupId>
```
adutra, line 42: I think it makes sense to instruct users to use the extension BOM.

quickstart/pom.xml
```diff
@@ -41,90 +38,110 @@
...
   <dependencies>
     <dependency>
-      <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-resteasy</artifactId>
+      <groupId>com.datastax.oss.quarkus</groupId>
+      <artifactId>cassandra-quarkus-client</artifactId>
     </dependency>
-    <!-- Tinkerpop uses JCL directly; redirect JCL calls to JBoss logging  -->
     <dependency>
-      <groupId>org.jboss.logging</groupId>
```
adutra, line 53: See my other comment; not exclusing jcl-over-slf4j makes this dependency useless.

quickstart/pom.xml
```diff
@@ -41,90 +38,110 @@
...
-      <groupId>com.datastax.oss</groupId>
-      <artifactId>java-driver-mapper-runtime</artifactId>
-      <version>${datastax-java-driver.version}</version>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-resteasy-mutiny</artifactId>
     </dependency>
     <dependency>
       <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-mutiny</artifactId>
+      <artifactId>quarkus-smallrye-health</artifactId>
```
adutra, line 77: Was missing. I think we need it in order for the health check to be available in the quickstart.

runtime/pom.xml
```diff
@@ -42,48 +42,17 @@
...
-      </exclusions>
-    </dependency>
-    <dependency>
-      <groupId>com.datastax.oss</groupId>
-      <artifactId>java-driver-mapper-runtime</artifactId>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-arc</artifactId>
     </dependency>
     <dependency>
-      <groupId>org.graalvm.nativeimage</groupId>
```
adutra, line 79: Not required, probably a leftover from the time where we had substitutions in the extension.

runtime/pom.xml
```diff
@@ -95,12 +64,13 @@
       <optional>true</optional>
     </dependency>
     <dependency>
-      <groupId>org.eclipse.microprofile.metrics</groupId>
```
adutra, line 98: Not required.

runtime/pom.xml
```diff
@@ -95,12 +64,13 @@
...
-      <artifactId>microprofile-metrics-api</artifactId>
+      <groupId>com.datastax.oss</groupId>
+      <artifactId>java-driver-core</artifactId>
     </dependency>
     <dependency>
-      <groupId>io.quarkus</groupId>
-      <artifactId>quarkus-mutiny</artifactId>
+      <groupId>com.datastax.oss</groupId>
+      <artifactId>java-driver-mapper-runtime</artifactId>
+      <optional>true</optional>
```
adutra, line 73: Making it optional. This effectively solves #82 .

bom-runtime/pom.xml
adutra: This effectively solves #79 .

tomekl007: see my comment here:
https://github.com/datastax/cassandra-quarkus/issues/79#issuecomment-646476395

deployment/pom.xml
tomekl007: Same situation, banned dependencies:
https://github.com/datastax/cassandra-quarkus/issues/79#issuecomment-646476395

quickstart/pom.xml
tomekl007: yes, but quickstart is not showing an example of healthcheck now (it is only in the integration-tests)

adutra: Do you mean that there is no health check available right now in the quickstart, or that the health check is available, but is not explained in the quickstart guide?

tomekl007: You are right, there is a section about health-checks in the quickstart guide:
`Connection Health Check` so it's ok to add this dependency.

bom-runtime/pom.xml
adutra: As I said in https://github.com/datastax/cassandra-quarkus/pull/77#issuecomment-646616843 , I think we are fine here. We don't need to (and should not) exclude JCL over SLF4J, and any other dependency other than Dropwizard.

adutra:
I'm going to merge this to check if the Ecosystem CI nightly build passes.

PR #91, opened 2020-06-19
issue84 Add a no-arg constructor to QuarkusCqlSessionBuilder

tomekl007:


runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/session/QuarkusCqlSessionBuilder.java
```diff
@@ -29,17 +29,23 @@
...
-
-  public QuarkusCqlSessionBuilder(
-      MetricRegistry metricRegistry,
-      EventLoopGroup mainEventLoop,
-      boolean useQuarkusNettyEventLoop) {
+  private MetricRegistry metricRegistry;
+  private EventLoopGroup mainEventLoop;
+  private boolean useQuarkusNettyEventLoop;
+
+  public QuarkusCqlSessionBuilder withMetricRegistry(MetricRegistry metricRegistry) {
```
adutra, line 36: Do you mind adding nullability annotations? It would be nice to be able to tell if all these items are required or not.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/session/QuarkusCqlSessionBuilder.java
```diff
@@ -33,12 +33,12 @@
...
   private boolean useQuarkusNettyEventLoop;
 
-  public QuarkusCqlSessionBuilder withMetricRegistry(MetricRegistry metricRegistry) {
+  public QuarkusCqlSessionBuilder withMetricRegistry(@NonNull MetricRegistry metricRegistry) {
     this.metricRegistry = metricRegistry;
     return this;
   }
 
-  public QuarkusCqlSessionBuilder withMainEventLoop(EventLoopGroup mainEventLoop) {
+  public QuarkusCqlSessionBuilder withMainEventLoop(@NonNull EventLoopGroup mainEventLoop) {
```
adutra, line 41: I'm not sure this is true. From what I see, `CassandraClientProducer#setMainEventLoop` is not always called, and as a consequence, we may pass a null reference here in `CassandraClientProducer#createCassandraClient`.

Thinking a bit about this, maybe we don't need the `useQuarkusNettyEventLoop` boolean at all. We could simply turn this parameter `@Nullable`, and if it is null, it means "don't use quarkus event loop". Wdyt?

tomekl007: I agree that is should be `@Nullable`, but regarding removing the `useQuarkusNettyEventLoop` this will still need to exist because it's a user-provided setting. 
I think that it is better to use this setting and be explicit:
```
  if (useQuarkusNettyEventLoop) {
      return new QuarkusNettyOptions(this, mainEventLoop, mainEventLoop);
    } else {
      return super.buildNettyOptions();
    }
```
wdyt?

adutra: I don't want to remove `useQuarkusNettyEventLoop` from `CassandraClientProducer`, just from `QuarkusCqlSessionBuilder` and `QuarkusDriverContext`. Because in these classes the 2 fields `mainEventLoop` and `useQuarkusNettyEventLoop` are tied together, that is, if `useQuarkusNettyEventLoop` is true then `mainEventLoop` is non-null and vice versa. So they are a bit redundant in these classes. You can rewrite `buildNettyOptions` very easily:

```
  @Override
  protected NettyOptions buildNettyOptions() {
    if (mainEventLoop != null) {
      return new QuarkusNettyOptions(this, mainEventLoop, mainEventLoop);
    } else {
      return super.buildNettyOptions();
    }
  }
```

BTW a better name for this field would be `quarkusEventLoop`.

PR #95, opened 2020-06-22
Add setting for DataStax Astra secure connect bundle (fixes #94)

### LINKED ISSUE #94: Add setting for DataStax Astra secure connect bundle

adutra:


runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientProducer.java
```diff
@@ -138,9 +138,15 @@ private void configureMetricsSettings(
...
-        DefaultDriverOption.LOAD_BALANCING_LOCAL_DATACENTER, config.localDatacenter);
+    config.contactPoints.ifPresent(
+        v -> configLoaderBuilder.withStringList(DefaultDriverOption.CONTACT_POINTS, v));
+    config.localDatacenter.ifPresent(
+        v ->
+            configLoaderBuilder.withString(DefaultDriverOption.LOAD_BALANCING_LOCAL_DATACENTER, v));
+    config.secureConnectBundle.ifPresent(
+        v ->
+            configLoaderBuilder.withString(
+                DefaultDriverOption.CLOUD_SECURE_CONNECT_BUNDLE, v.toAbsolutePath().toString()));
```
absurdfarce, line 149: One (perhaps) minor suggestion: do we want to key this off of the value of config.secureConnectBundle instead?  Something along the lines of:

```
if secureConnectBundle
   set connectBundle prop
else
   set contactPoints prop
   set localDC prop
```

That might make the interaction between these properties a bit more explicit.  I don't know that it's _necessary_; what you have here should certainly work.

adutra: That was my initial design, but I eventually changed it on the grounds that this is not the most appropriate location to decide which properties need to be set and which shouldn't. At this level we should just propagate whatever is defined. One of the reasons for this is: these settings need to be merged with others defined in application.conf; so conflicts between incompatible settings could happen even if we do here as you suggest (for example, if local datacenter is defined here, and a secure bundle in application.conf).

absurdfarce: Specifically I was thinking of something along the lines of:

```
    config.secureConnectBundle.map(v ->
            configLoaderBuilder.withString(
                    DefaultDriverOption.CLOUD_SECURE_CONNECT_BUNDLE, v.toAbsolutePath().toString())
    ).orElseGet(() -> {

      Preconditions.checkArgument(config.contactPoints.isPresent(),"Contact points must be provided if not using Astra")
      Preconditions.checkArgument(config.localDatacenter.isPresent(),"Local DC must be provided if not using Astra")
      configLoaderBuilder.withStringList(DefaultDriverOption.CONTACT_POINTS, config.contactPoints.get());
      configLoaderBuilder.withString(DefaultDriverOption.LOAD_BALANCING_LOCAL_DATACENTER, config.localDatacenter.get());
      return configLoaderBuilder;
    });
```

I'm not 💯 convinced this is cleaner code even though it makes the expected relationships between those options cleaner.  There's a mixing of expressions and statements here that I don't... love; it's certainly arguable that this approach would be cleaner if we'd avoided void methods from the beginning but we have them now.

Hardly required; you are free to use or not use at your discretion :)

absurdfarce: Ah, apologies @adutra , I missed your reply.

Are you envisioning resolving these issues in... CassandraClientConnectionConfig then?

adutra: > Or perhaps just push these settings along to the driver and let the driver sort it out?

Yes, that. The driver already has all the logic to warn/fail on incompatible settings. And besides, this logic is about to change, see [JAVA-2813](https://datastax-oss.atlassian.net/browse/JAVA-2813).

Issue #46, opened 2020-05-29
[JAVA-2697] Map all basic and important settings into quarkus settings mechanism (ssl, cloud) and improve config mechanism


adutra:
<p>Imported from <a href="https://datastax-oss.atlassian.net/browse/JAVA-2697">JAVA-2697</a>, originally reported by @tomekl007 on Thu, 19 Mar 2020 09:09:44 GMT.</p>
<hr />
<p><a href="https://datastax-oss.atlassian.net/browse/JAVA-2682" rel="nofollow">https://datastax-oss.atlassian.net/browse/JAVA-2682</a> implemented mechanism for propagating settings from quarkus to java-driver.</p> 
<p>In this ticket. we should go through all the required settings and map them in a similar way.</p> 
<p>We should map only important settings like SSL, cloud. The user still can use application.conf to configure all underlying java-driver settings.</p> 
<p>See docs about including resources:</p> 
<p><a href="https://quarkus.io/guides/writing-native-applications-tips" rel="nofollow">https://quarkus.io/guides/writing-native-applications-tips</a></p> 
<p>We should decide which settings can be moved to deployment (they will be static once constructed so cannot be reload on the runtime)</p>

adutra:
<p>Imported from <a href="https://datastax-oss.atlassian.net/browse/JAVA-2697?focusedCommentId=52400">JAVA-2697</a>, originally commented by @tomekl007 on Thu, 26 Mar 2020 07:52:59 GMT.</p>
<hr />
<p>There is a discussion about settings:</p> 
<p><a href="https://github.com/quarkusio/quarkus/issues/7876#issuecomment-603773398" rel="nofollow">https://github.com/quarkusio/quarkus/issues/7876#issuecomment-603773398</a> </p> 
<blockquote>
 <p>So look at the runtime configs you have and think whether you could prepare the driver at build time and load less class at runtime time if you could freeze the config to build time immutable. And then look how loud people complain. Shake it and make your choice <br> We tend to put stuff aggressively at build time and backup as needed.</p> 
 <p>I don't know the extend of the Cassandra config but:</p> 
 <ul> 
  <li>IP, url, username, password and definitely runtime</li> 
 </ul> 
 <ul> 
  <li>loadbanancer policy, cypher suite, etc (involving a piece of code switch in the driver) could be made build time</li> 
 </ul> 
 <p>We can go over them specifically.</p> 
 <p>Speaking of properties, we have done the following in a few frameworks:</p> 
[...]

PR #107, opened 2020-06-30
Add more configuration settings (fixes #72, fixes #105)

### LINKED ISSUE #72: Add config settings for keyspace and graph name
Just as a feedback for your actual work. I'm using the snapshot of the quarkus-cassandra driver and use it for graph statements. It is actually working, but I have to set the graph name on every statement, cause the property quarkus.cassandra.graph.name is ignored. Such "Standart" properties should be supported.

### LINKED ISSUE #105: Expose advanced.reconnect-on-init and advanced.resolve-contact-points in CassandraClientConfig
The driver option `advanced.reconnect-on-init` is false by default, but it probably makes sense, for microservices architectures to set it to true by default.

The driver option `advanced.resolve-contact-points` is also true by default, but it probably makes sense to set it to false instead.

We could expose it in `CassandraClientConfig` for easier access.

adutra:


adutra:
@thomasbecker just a heads up that, if this PR is accepted, the settings for authentication will change to:

```
quarkus.cassandra.auth.username = <username>
quarkus.cassandra.auth.password = <password>
```

adutra:
@UnvirtualHH this PR solves #72 so you will have the option to set the graph name in application.properties - if the PR is accepted as is.

adutra:
@tomekl007 after this PR we still have #46 open, we can use that issue to add more authentication options and more importantly, SSL.

UnvirtualHH:
I have successfully tested the graph configuration properties. All working fine.

PR #108, opened 2020-07-05
Simplify quickstart, use session keyspace in daos

adutra:
I think parameterizing daos by keyspace or table are a more advanced topic, I suggest that we keep the quickstart really simple by reusing the session keyspace.


quickstart/src/test/resources/application.properties
```diff
@@ -18,4 +18,4 @@
 # used for tests; see CassandraTestResource in the test-framework module.
 quarkus.cassandra.contact-points=${quarkus.cassandra.docker_host}:${quarkus.cassandra.docker_port}
 quarkus.cassandra.local-datacenter=datacenter1
-fruit.dao.keyspace=k1
```
tomekl007, line 21: The idea behind it was that the user may want to create a DAO per keyspace:
https://github.com/datastax/cassandra-quarkus/pull/108/files#diff-f6522f26502dbc8b454e6a91a7e59436L35
I wanted to should the example of how to do it in the client code.
With the current change, the user will need to share the same keyspace for all DAO's. 
I don't know if that a problem.

adutra: I think that now that we can specify the keyspace in `CassandraClientConfig`, having another keyspace specified elsewhere will create confusion. We should promote the good practice of binding to one keyspace for the entire app imo.

Issue #125, opened 2020-08-25
Don't call close() if session wasn't initialized


adutra:
We add a shutdown task to close the session, see [here](https://github.com/datastax/cassandra-quarkus/blob/bde8d2bccdd39d058b42542e98348061532bef8c/runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientRecorder.java#L44). 

I suspect that calling `close()` on an uninitialized proxy triggers the whole initialization procedure. This is a corner case for sure, but someone hit that already.

_Originally posted by @adutra in https://github.com/datastax/cassandra-quarkus/issues/124#issuecomment-680020784_

tomekl007:
https://github.com/datastax/cassandra-quarkus/issues/124#issuecomment-679997511
@JiriOndrusek could I see the code and tests that are executed to debug this problem?

JiriOndrusek:
Hi @tomekl007, there is my work in progress in following branch https://github.com/JiriOndrusek/camel-quarkus/tree/for_datastax_125

To execute test, use `mvn clean install -f integration-tests/cassandraql/`
You should see empty test passing successfully and then it will hang up and wait for the connection to cassandra (this issue)

tomekl007:
I am debugging the `close()` behavior and confirmed that when `eager-session-init=false` and the `QuarkusCqlSession` is not injected anywhere in the application, the close() method is triggering the `CassandraClientProducer#createCassandraClient()`:
```
[...]

PR #126, opened 2020-08-28
issue-125: Don't call close() if session wasn't initialized

tomekl007:


tomekl007:
@JiriOndrusek 
could you validate that this fixes your problem?

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientStarter.java
```diff
@@ -36,10 +36,14 @@
...
-      @Observes StartupEvent event, QuarkusCqlSession session, CassandraClientConfig config) {
+      @Observes StartupEvent event,
+      QuarkusCqlSession sessionProxy,
+      CassandraClientConfig config,
+      QuarkusCqlSessionState quarkusCqlSessionState) {
     if (config.cassandraClientInitConfig.eagerSessionInit) {
       LOG.debug("Triggering eager initialization of Quarkus session at startup");
-      session.getName();
+      sessionProxy.getName();
+      quarkusCqlSessionState.setInitialized();
```
adutra, line 46: I think this is redundant, you already call it in `createCassandraClient` above.

tomekl007: you are right, removed

JiriOndrusek:
@tomekl007 
I've tried to execute the test and it works, thanks.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientProducer.java
```diff
@@ -51,7 +51,15 @@
...
-  public QuarkusCqlSession createCassandraClient() {
+  public QuarkusCqlSessionState quarkusCqlSessionState() {
+    return QuarkusCqlSessionState.notInitialized();
+  }
+
+  @Produces
+  @ApplicationScoped
+  @Unremovable
+  public QuarkusCqlSession createCassandraClient(QuarkusCqlSessionState quarkusCqlSessionState) {
+    quarkusCqlSessionState.setInitialized();
```
adutra, line 62: It would probably be more correct if this was the last statement in this method, right before `return session;`.

adutra (approved):

adutra:
Take care with commit messages, the following would be more appropriate as per the contribution guidelines: `Don't call close() if session wasn't initialized (fixes #125)`.

Issue #23, opened 2020-05-29
[JAVA-2786] Extend health check to all connected nodes


adutra:
<p>Imported from <a href="https://datastax-oss.atlassian.net/browse/JAVA-2786">JAVA-2786</a>, originally reported by @adutra on Tue, 19 May 2020 16:07:15 GMT.</p>
<hr />
<p>As decided internally with other drivers, we should implement health checks following the definition below:</p> 
<blockquote>
 <p>A cluster is healthy from the application standpoint if the driver has at least one open connection to all the nodes not in IGNORED state.</p>
</blockquote> 
<p>This could be done with session.getMetadata().getNodes() + Node.getOpenConnections() but it’s probably better to wait for <span> <a href="https://datastax-oss.atlassian.net/browse/JAVA-2693" rel="nofollow">  JAVA-2693 </a> <span>Resolved</span> </span> .</p>

adutra:
<p>Link imported from Jira: this issue (<a href="https://datastax-oss.atlassian.net/browse/JAVA-2786">JAVA-2786</a>) --- <em>is blocked by</em> ---> <a href="https://datastax-oss.atlassian.net/browse/JAVA-2742">JAVA-2742</a>.</p>

adutra:
<p>Link imported from Jira: this issue (<a href="https://datastax-oss.atlassian.net/browse/JAVA-2786">JAVA-2786</a>) --- <em>relates to</em> ---> <a href="https://datastax-oss.atlassian.net/browse/JAVA-2693">JAVA-2693</a> (<a href="https://github.com/datastax/cassandra-quarkus/issues/49">issue #49</a>).</p>

PR #127, opened 2020-08-31
Provide a lazy wrapper for QuarkusCqlSession (fixes #106)

### LINKED ISSUE #106: Provide a lazy wrapper for QuarkusCqlSession
Instead of injecting `CompletionStage<QuarkusCqlSession>` we could provide a session wrapper so that users can inject `QuarkusCqlSession` directly and avoid blocking Vertx threads even if they opt out from eager init at startup.

tomekl007:
todo: 
- [x] migrate Mapper usages to CompletionStage<Mapper> (same for Daos)
- [x] migrate CompletionStage<Uni<T>> to `Unit<T>` (same for multi)
- [x] reimplement `CassandraHealthCheck` to work in a non-blocking way (blocked on https://github.com/datastax/cassandra-quarkus/issues/128)
- [x] rework `PagingIterable<Product> findAll();` to an async version
- [x] docs


runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientProducer.java
```diff
@@ -58,7 +61,8 @@ public QuarkusCqlSessionState quarkusCqlSessionState() {
   @Produces
   @ApplicationScoped
   @Unremovable
-  public QuarkusCqlSession createCassandraClient(QuarkusCqlSessionState quarkusCqlSessionState) {
+  public CompletionStage<QuarkusCqlSession> createCompletionStageOfCassandraClient(
```
tomekl007, line 64: It seems that injecting of the CompletionStage works without a problem, I think that we need to document 4 cases:
1. client's code is setting `eager-session-init=false` and injecting the only CompletionStage<QuarkusCqlSession> - in that case, there should be no problem with blocking on Vert.x thread.
2. client's code is setting `eager-session-init=false` and injecting `QuarkusCqlSession` - this is problematic and should be avoided. We can detect that and log a warning, see https://github.com/datastax/cassandra-quarkus/pull/127/files#diff-535ef3e00080110cf5c9d6d562874ff1R92-R94
3.  client's code is setting `eager-session-init=true` - injecting `QuarkusCqlSession` AND `CompletionStage<QuarkusCqlSession>` is allowed

tomekl007:
@adutra regarding your comment:
https://github.com/datastax/cassandra-quarkus/issues/106#issuecomment-681925654
I've tried to rework all the code in integraiton-tests to leverage `CompletionStage<QuarkusCqlSession>`: https://github.com/datastax/cassandra-quarkus/pull/127/commits/9ddf9535d8850239b0f499fc3f563f9a98ef4912
I was able to make it work, ITs are passing. The mapper, DAOs, and services are working in a non-blocking way.
One thing that I still need to fix is:
https://github.com/datastax/cassandra-quarkus/pull/127/files#diff-53e34354ebbb936d252e50949e04ea37R35-R38
- flattening `CompletionStage<Uni<Void>>` to `Uni<Void>` feel free to suggest something. 

Unfortunately, we have a different problem when `eager-session-init=false` and injecting `QuarkusCqlSession`, that is `HealhCheck` API used in the quarkus framework. If you will look at the `CassandraHealthCheck` you can see the health-check method MUST return `HealthCheckResponse`, and cannot return `CompletionStage<HealthCheckResponse>`
```
 @Override
  public HealthCheckResponse call() 
```
IOW the quarkus health check API is blocking and cannot be implemented in the async. Due to that fact, we need to inject `QuarkusCqlSession` or block on `CompletionStage<QuarkusCqlSession>` and it may lead to the blocking behavior that we wanted to avoid. I am not sure if this is feasible to alter the `org.eclipse.microprofile.health.HealthCheck` to expose the reactive version of the `call()` method. Maybe we could create a dedicated `ThreadPool` inside the `CassandraHealthCheck` and block on this thread pool? (instead of Vert.x thread pool)?

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
```diff
@@ -276,9 +276,8 @@ CassandraClientBuildItem cassandraClient(
   @BuildStep
   HealthBuildItem addHealthCheck(CassandraClientBuildTimeConfig buildTimeConfig) {
     return new HealthBuildItem(
-        "com.datastax.oss.quarkus.runtime.internal.health.CassandraHealthCheck",
-        buildTimeConfig.healthEnabled,
-        "cassandra");
+        "com.datastax.oss.quarkus.runtime.internal.health.CassandraAsyncHealthCheck",
```
tomekl007, line 279: It seems that it is safe to switch to Async implementation. It has a dependency on `Uni`, but this type is always present in the `cassandra-quarkus-deployment` module:
```
    <dependency>
      <groupId>io.quarkus</groupId>
      <artifactId>quarkus-mutiny-deployment</artifactId>
    </dependency>
```

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/health/CassandraAsyncHealthCheck.java
```diff
@@ -0,0 +1,104 @@
...
+@Readiness
+@ApplicationScoped
+public class CassandraAsyncHealthCheck implements AsyncHealthCheck {
+  /** Name of the health-check. */
+  private static final String HEALTH_CHECK_NAME = "DataStax Apache Cassandra Driver health check";
+
+  private static final Type COMPLETION_STAGE_OF_QUARKUS_CQL_SESSION_TYPE =
+      new TypeLiteral<CompletionStage<QuarkusCqlSession>>() {}.getType();
+  static final String HEALTH_CHECK_QUERY =
+      "SELECT data_center, release_version, cluster_name, cql_version FROM system.local";
```
adutra, line 44: Should we take advantage of this PR to change this to what we did for Spring Boot?

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/health/CassandraHealthCheck.java
```diff
@@ -21,14 +21,12 @@
...
 import javax.annotation.PostConstruct;
-import javax.enterprise.context.ApplicationScoped;
 import org.eclipse.microprofile.health.HealthCheck;
 import org.eclipse.microprofile.health.HealthCheckResponse;
 import org.eclipse.microprofile.health.HealthCheckResponseBuilder;
-import org.eclipse.microprofile.health.Readiness;
 
-@Readiness
-@ApplicationScoped
+// @Readiness
```
adutra, line 28: Is this temporary? Should we remove this class?

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientProducer.java
adutra: 1. I think a warning is a bit too much. This is not always a problem, it's only a problem afaict if you use Vert.x routes. But if you use regular REST controllers, it's not a problem.
2. Should we make eager-session-init false by default?

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientStarter.java
```diff
@@ -36,10 +38,13 @@
...
   public void startup(
-      @Observes StartupEvent event, QuarkusCqlSession sessionProxy, CassandraClientConfig config) {
+      @Observes StartupEvent event,
+      CompletionStage<QuarkusCqlSession> sessionProxy,
+      CassandraClientConfig config)
+      throws ExecutionException, InterruptedException {
     if (config.cassandraClientInitConfig.eagerSessionInit) {
       LOG.debug("Triggering eager initialization of Quarkus session at startup");
-      sessionProxy.getName();
+      sessionProxy.toCompletableFuture().get().getName();
```
adutra, line 47: I think you don't need `.getName()` anymore.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
absurdfarce: This _seems_ right.  How would this come into play with transitive exclusions thought?  I'm thinking of something like a customer who doesn't want to write reactive code in their app and thus excludes Mutiny as a transitive dependency... and then that customer unexpectedly loses access to health checks.

absurdfarce: Hmmm, that's probably a non-issue.  Forgot QuarkusCqlSession already extends a number of interfaces which expose Uni/Multi directly... so I suspect you'd run aground well before you go to this point.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/health/CassandraAsyncHealthCheck.java
tomekl007: I was thinking about it but this PR is already introducing a lot of new stuff, so I prefer to do it separately

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/health/CassandraHealthCheck.java
tomekl007: I think that we can remove this health-check entirely. Done in https://github.com/datastax/cassandra-quarkus/pull/127/commits/5a31d0cb19096909e9e0837493c719f3867a7bde

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
tomekl007: yes, I was trying to make it NOT work by excluding mutiny, and reactive dependencies in the `integration-tests` and `quickstart` and it was working properly. So it seems that it is safe to make `CassandraAsyncHealthCheck` default

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientProducer.java
tomekl007: 1. You are right, we would need to detect if someone is using vert.x. I will change it to an `info` level
2. I doubt that majority of quarkus users are using everything async.  I think that they will pick the `QuarkusCqlSession` by default. So maybe we could leave it as `true`, but mention in the docs that if they want to make they apps fully async, then they should set `eager-session-init = false` and inject `CompletionStage<QuarkusCqlSession>`, wdyt?

integration-tests/src/main/java/com/datastax/oss/quarkus/tests/mapper/InventoryMapper.java
```diff
@@ -22,16 +22,17 @@
...
-  ProductDao productDao(@DaoKeyspace CqlIdentifier keyspace);
+  CompletionStage<ProductDao> productDao(@DaoKeyspace CqlIdentifier keyspace);
 
   @DaoFactory
-  ProductReactiveDao productReactiveDao(@DaoKeyspace CqlIdentifier keyspace);
+  CompletionStage<ProductReactiveDao> productReactiveDao(@DaoKeyspace CqlIdentifier keyspace);
 
   @DaoFactory
-  CustomerDao customerDao(@DaoKeyspace CqlIdentifier keyspace);
+  CompletionStage<CustomerDao> customerDao(@DaoKeyspace CqlIdentifier keyspace);
```
absurdfarce, line 37: Something about this API has been bothering me since I first saw it but I'm still having trouble articulating exactly what my concern is.  All of the following are mixed together in there:

* Do we really want to expose CompletionStage at this level of the stack?  I completely understand the need to change it at lower levels but it's not immediately clear to me it should surface this high up
* This forces users to interact with the async API for _every_ operation; I'm not sure there's enough benefit there to warrant the intrusion
* After the stage has actually been completed there isn't a huge amount of value in the combination functions offered by CompletionStage.  Since you now have a value (rather than the promise of a future one which may or may not be present) you can manipulate it in any number of ways to achieve similar effects.

So what would the alternatives be?

* You could return the raw value (as was done in the original API) and block if it isn't available yet.  This isn't... great, but will possibly be applicable for a wide number of use cases
** I can easily imagine customers adding blocking methods of these API calls to do just that
* Return a CompletableFuture (rather than a CompletionStage) to offer at least some level of support for both models (blocking vs. async + callback) in a more intuitive way.

I get that this is only a test and that users may implement this differently.  I guess I'm working off of the idea that tests in your repo often serve as a model for how to do things... so I'm wanting to put forward the best example here.

If I'm the only one concerned about this I won't belabor the point, and I certainly don't want to hold up the PR just for this; there's lots of other good stuff in here.  But it did seem worth raising.

integration-tests/src/main/java/com/datastax/oss/quarkus/tests/mapper/InventoryMapperProducer.java
```diff
@@ -19,45 +19,60 @@
...
-        .orElseThrow(() -> new IllegalStateException("Session is not bound to a keyspace"));
+  CompletionStage<CustomerDao> produceCustomerDao() {
+    return sessionCompletionStage
+        .thenApply(
+            s ->
+                s.getKeyspace()
+                    .orElseThrow(
+                        () -> new IllegalStateException("Session is not bound to a keyspace")))
+        .thenCompose(
+            keyspace -> mapperCompletionStage.thenCompose(mapper -> mapper.customerDao(keyspace)));
```
absurdfarce, line 76: We've definitely lost some concision in the code here.  I was trying to get back to something closer to the original code with the following:

```java
-        .thenCompose(
-            keyspace -> mapperCompletionStage.thenCompose(mapper -> mapper.customerDao(keyspace)));
+        .thenCombine(mapperCompletionStage, (keyspace, mapper) -> mapper.customerDao(keyspace));
```

This fails because mapper.customerDao() now returns a ComplationStage<CustomerDao> rather than CustomerDao itself.  I mention this mainly in case anybody has a cleaner way to get at what I was trying to get at above... the nested thenCompose() calls are (primarily) what I was trying to avoid.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/health/CassandraAsyncHealthCheck.java
```diff
@@ -0,0 +1,104 @@
...
+                        .executeAsync(HEALTH_CHECK_QUERY)
+                        .thenApply(AsyncPagingIterable::one)
+                        .thenApply(
+                            result -> {
+                              HealthCheckResponseBuilder builder =
+                                  HealthCheckResponse.named(HEALTH_CHECK_NAME).up();
+                              if (result == null) {
+                                return builder
+                                    .down()
+                                    .withData("reason", "system.local returned null")
```
absurdfarce, line 75: Perhaps this should be "Query of system.local returned null"?  Or maybe just "Health check query returned null"?

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/health/CassandraAsyncHealthCheck.java
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  private Map<String, String> extractInfoFromResult(Row result) {
+    HashMap<String, String> details = new HashMap<>();
+    details.put("datacenter", result.getString("data_center"));
+    details.put("releaseVersion", result.getString("release_version"));
+    details.put("clusterName", result.getString("cluster_name"));
+    details.put("cqlVersion", result.getString("cql_version"));
+    return details;
+  }
```
absurdfarce, line 103: Nit: it's probably slightly better to specify the list of columns we wish to query and extract in a collection somewhere.  We can then use that collection to build HEALTH_CHECK_QUERY as well as iterate over the same collection here when we wish to extract values.  Advantage is that we specify those columns in a single spot and re-use them in both operations.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientProducer.java
```diff
@@ -68,8 +73,29 @@ public QuarkusCqlSession createCassandraClient(QuarkusCqlSessionState quarkusCql
...
+  @ApplicationScoped
+  @Unremovable
+  public QuarkusCqlSession createCassandraClient(CompletionStage<QuarkusCqlSession> sessionFuture) {
+    LOG.trace("Produce QuarkusCqlSession bean.");
+    if (!config.cassandraClientInitConfig.eagerSessionInit) {
+      LOG.info(
+          "Injecting QuarkusCqlSession and setting eager-session-init = false may cause deadlock on the Vert.x thread. "
+              + "Please set it to true, inject only CompletionStage<QuarkusCqlSession>, "
+              + "or assert that initializing of the QuarkusCqlSession in your Application is happening on your thread pool (not Vert.x event loop).");
+    }
```
absurdfarce, line 97: To the point by @adutra above... this is only true for Vert.x reactive routes (since those are the only ones that run on the Vert.x thread pool).  Shouldn't we at least mention that constraint in the message?  Ideally we'd actually test for the usage of reactive routes in the app and only warn if they're actually in use... but failing that mentioning them explicitly seems like a good idea.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientProducer.java
absurdfarce: I agree with @tomekl007 here (leave the default as eager-session-init = true), but for a _slightly_ different reason.  eager-session-init = false is the one case that's known to produce unreliable results (for Vert.x reactive routes).  There's a decent argument that we should be more defensive by default and avoid this configuration by default while still allowing users to access it easily if they so desire.

integration-tests/src/main/java/com/datastax/oss/quarkus/tests/mapper/InventoryMapperProducer.java
tomekl007: As far as I know, the `thenCompose` pattern is recommended for achieving this goal using `CompletionStage` API

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/health/CassandraAsyncHealthCheck.java
tomekl007: This is copy from older health check and will be changed in the https://github.com/datastax/cassandra-quarkus/issues/23 so not worth changing now

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/health/CassandraAsyncHealthCheck.java
tomekl007: same as here:
https://github.com/datastax/cassandra-quarkus/pull/127#discussion_r482764654

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientProducer.java
tomekl007: I think that this part: `or assert that initializing of the QuarkusCqlSession in your Application is happening not on the Vert.x event loop.` consists of the case when the user is not using Vert.x - the end-user can assert that initialization is happening not on Vert.x event loop

absurdfarce: Yeah, _I_ understand that those things are connected... but it's not immediately clear to me that a general Quarkus user will connect "something happening on a vert.x event loop" to "I'm using reactive routes".  Regardless it's not a big concern.

integration-tests/src/main/java/com/datastax/oss/quarkus/tests/mapper/InventoryMapper.java
absurdfarce: I'm still skeptical, but at this point we're debating between returning the CompletionStage or the future you get from it... and that's certainly no reason to hold this up.

absurdfarce (approved):

PR #130, opened 2020-09-01
Use a composite loader to inject Quarkus properties into the driver configuration (fixes #75)

### LINKED ISSUE #75: Use a composite loader to inject Quarkus properties into the driver configuration
Currently the extension uses the programmatic config loader to inject the options coming from the Quarkus config (`CassandraClientConfig` and `CassandraClientConnectionConfig`).

The problem is that this doesn't allow users to bring their own config loader: if they want to customize low-level options via the native driver config, it has to be an `application.conf` file in the classpath, it's not possible to use other built-in loader (e.g. `DriverConfigLoader.fromUrl`), or a custom loader.

Driver 4.7.0 introduces `DriverConfigLoader.compose()` to solve this:
- use a map-based config loader for the Quarkus options:
    ```java
    OptionsMap quarkusConfig = new OptionsMap();

    // copy all the options of CassandraClientConfig, etc, into the map
    ```
- override methods in `QuarkusCqlSessionBuilder` to always use the Quarkus options first:
    ```java
    @NonNull
    public QuarkusCqlSessionBuilder withConfigLoader(@Nullable DriverConfigLoader configLoader) {
      return super.withConfigLoader(overrideWithQuarkusConfig(configLoader));
    }
  
    @NonNull
    @Deprecated
    protected DriverConfigLoader defaultConfigLoader() {
      return overrideWithQuarkusConfig(super.defaultConfigLoader());
    }
  
    @NonNull
    protected DriverConfigLoader defaultConfigLoader(@Nullable ClassLoader classLoader) {
      return overrideWithQuarkusConfig(super.defaultConfigLoader(classLoader));
    }

    private DriverConfigLoader overrideWithQuarkusConfig(DriverConfigLoader configLoader) {
      return DriverConfigLoader.compose(
          DriverConfigLoader.fromMap(quarkusConfig), // takes precedence
          configLoader);
    }
    ```
- we should also add a hook to allow the user to inject their own config loader. I'm not sure of the best way to do that in Quarkus, maybe make it injectable into the session producer bean?

PS: don't use `OptionsMap.driverDefaults()` to initialize the Quarkus config, we want it to only contain the options explicitly declared by Quarkus.

tomekl007:


runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/session/QuarkusCqlSessionBuilder.java
```diff
@@ -43,11 +45,38 @@ public QuarkusCqlSessionBuilder withQuarkusEventLoop(@Nullable EventLoopGroup qu
...
+    return overrideWithQuarkusConfig(super.defaultConfigLoader());
+  }
+
+  @NonNull
+  protected DriverConfigLoader defaultConfigLoader(@Nullable ClassLoader classLoader) {
+    return overrideWithQuarkusConfig(super.defaultConfigLoader(classLoader));
+  }
+
+  private DriverConfigLoader overrideWithQuarkusConfig(DriverConfigLoader configLoader) {
+    return DriverConfigLoader.compose(
```
adutra, line 75: I suggest that we revert this portion until JAVA-2887 is fixed.

adutra:
Rebased and force-pushed to update to latest `master`.

adutra:
Closing due to inactivity.

PR #137, opened 2020-10-01
Use driver's own MicroProfileMetricsFactory

adutra:


runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientProducer.java
```diff
@@ -162,13 +161,20 @@ private void configureProtocolCompression(
...
+      configLoaderBuilder.withStringList(
+          DefaultDriverOption.METRICS_NODE_ENABLED, enabledNodeMetrics);
+      configLoaderBuilder.withStringList(
+          DefaultDriverOption.METRICS_SESSION_ENABLED, enabledSessionMetrics);
+    }
   }
 
   private void configureRuntimeSettings(ProgrammaticDriverConfigLoaderBuilder configLoaderBuilder) {
+    // FIXME the name is required for metrics to have the expected prefix
+    configLoaderBuilder.withString(DefaultDriverOption.SESSION_NAME, "cassandra");
```
adutra, line 177: I'm uncomfortable about this. I think we need to investigate a way to augment metric names with an arbitrary prefix in the driver.

PR #149, opened 2021-01-26
Quickstart enhancements

### LINKED ISSUE #114: Place quickstart guide inside quickstart module
We know now how to publish the quickstart guide to the main Quarkus repo. We don't need to generate PDF or HTML, only create a PR [here](https://github.com/quarkusio/quarkus/tree/master/docs/src/main/asciidoc).

In this context it doesn't make sense to keep a separate `documentation` module. It would be simpler for users browsing this repo if the quickstart guide was the main README file for the `quickstart` module.

adutra:


quickstart/README.adoc
```diff
@@ -1,45 +1,395 @@
...
+TIP: If you don't know the name of your local datacenter, this value can be found by running the
+following CQL query: `SELECT data_center FROM system.local`.
+
+If you want to use Docker to run a Cassandra database, you can use the following command to launch
+one:
 
 [source,shell]
 ----
-docker run --name local-cassandra-instance cassandra
+docker run --name local-cassandra-instance -p 9042:9042 cassandra
```
jdonenine, line 357: Might be worth suggesting running the container with the `-d` option to keep it in the background.  Mainly just because in the next step we suggest a few other docker comments to setup the DB, which means starting up a new console.

quickstart/README.adoc
```diff
@@ -50,30 +400,321 @@ To retrieve fruits:
...
+When metrics are properly enabled, metric reports for all enabled metrics are available at the
+`/metrics` REST endpoint of your application.
+
+Running in dev mode with `mvn clean quarkus:dev`, if you point your browser to
+http://localhost:8080/metrics you should see a list of metrics; search for metrics whose names start
+with `vendor_cassandra`.
+
 == Running in native mode
 
 You can compile the application into a native binary using:
```
jdonenine, line 720: This was my first time ever using GraalVM and it's native compilation, I didn't expect the time that it took to complete the native build.  Might be worth mentioning here that native compilation can take a significant amount of time.

I was running inside of a lower-end VM though, so potentially it's faster on a higher end machine.

jdonenine (approved):
Hey, @adutra I went through the guide and was able to run everything as described, really good content here.  Great examples, guidance, and tips along the way.  I added just two minor comments about my experience going through the guide.  Overall, was very smooth - well once I used the right branch of code anyways 🤦 !

adutra:
Thank you @jdonenine for the review!

PR #205, opened 2022-01-28
Define Assertj version

gsmet:
Assertj version is not enforced by the Quarkus BOM anymore.

Due to some repeated binary compatibility issues, we had to remove it from the BOM.

Any chance you could release a new version with this fix before February 1st? We will need it for the 2.7 Quarkus Platform release.

### DIFF
--- a/deployment/pom.xml
+++ b/deployment/pom.xml
@@ -41,6 +41,11 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>org.assertj</groupId>
+        <artifactId>assertj-core</artifactId>
+        <version>${assertj.version}</version>
+      </dependency>
     </dependencies>
   </dependencyManagement>
   <dependencies>
--- a/integration-tests/pom.xml
+++ b/integration-tests/pom.xml
@@ -53,6 +53,11 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>org.assertj</groupId>
+        <artifactId>assertj-core</artifactId>
+        <version>${assertj.version}</version>
+      </dependency>
     </dependencies>
   </dependencyManagement>
   <build>
--- a/mapper-processor/pom.xml
+++ b/mapper-processor/pom.xml
@@ -42,6 +42,11 @@
         <type>pom</type>
         <scope>import</scope>
       </dependency>
+      <dependency>
+        <groupId>org.assertj</groupId>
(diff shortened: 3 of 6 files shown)


gsmet:
Ah, wait, no sorry, we don't need a release right now, we will need one for 2.8. But it would be nice to merge it ASAP so that we can have the Ecosystem CI green.

gsmet:
Still probably a good idea to do a release soon so that we don't forget about it :).

adutra (changes requested):
Thank you @gsmet we will look into getting a new release out soonish!

I think `runtime/.pom.xml.swp` was committed by mistake, could you please remove it?

gsmet:
> I think runtime/.pom.xml.swp was committed by mistake, could you please remove it?

Raaah. Will do, sorry about that.

gsmet:
Done. That's what I get for doing things in batches.

adutra:
> Done. That's what I get for doing things in batches.

No worries at all, I love when people report issues AND provide the fix at the same time :-)

Just one general remark: wouldn't it be simpler to declare AssertJ only in the parent pom and in the quickstart pom? The parent pom could look like this:

```xml
  <properties>
    <assertj.version>3.22.0</assertj.version>
  </properties>
  <dependencyManagement>
    <dependencies>
      <dependency>
        <groupId>org.assertj</groupId>
        <artifactId>assertj-core</artifactId>
        <version>${assertj.version}</version>
      </dependency>
    </dependencies>
  </dependencyManagement>
```

As far as I can tell, this produces the same effect as declaring AssertJ in all modules that use it..

gsmet:
No. Because if we do that, it would be part of your BOM and we don't want that. Keep in mind that the BOM inherits any dependency management coming from the parents (we had some very bad surprises in Quarkus with various components bringing a ton of unrelated things with their BOM).

What we do for Quarkus projects is that we have a `build-parent/pom.xml` with this sort of things and the BOM is explicitly not using this one.

adutra (approved):

gsmet:
Any chance this could be merged soon? Working on getting our Ecosystem CI green.

adutra:
Sure thing, sorry for the delay!

adutra:
> Ah, wait, no sorry, we don't need a release right now, we will need one for 2.8. But it would be nice to merge it ASAP so that we can have the Ecosystem CI green.

@gsmet what's the ETA for Quarkus 2.8? As in, when to you need us to release? :-)

gsmet:
If you release one month from now, that should be OK.

PR #215, opened 2022-12-10
Make DAO initialization during startup with eager init (fixes #214)

### LINKED ISSUE #214: DAO initialization is not happening on main thread with eager init
When eager init is on, the session and the mappers are all initialized during application startup, however DAO bean are still being produced in a lazy fashion.

┆Issue is synchronized with this [Jira Task](https://datastax-oss.atlassian.net/browse/QUAR-115) by [Unito](https://www.unito.io)

adutra:


runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientStarter.java
```diff
@@ -64,33 +66,39 @@ public void onStartup(@Observes StartupEvent event)
...
-        } else if (bean instanceof Uni) {
-          ((Uni<?>) bean).await().atMost(timeout);
-        }
-      } catch (TimeoutException | io.smallrye.mutiny.TimeoutException e) {
-        LOG.warn(
-            "Eager initialization of a {} bean did not complete within {}; "
-                + "resuming application startup with uninitialized bean.",
-            beanName,
-            timeout);
+      ClientProxy.unwrap(bean);
```
adutra, line 97: This was the missing bit, as not all DAO beans are completion stages or Unis.

PR #222, opened 2023-02-27
Adapt to Quarkus 3

adutra:
This PR aims to make the extension compatible with Quarkus 3.

The main pain point was the change in Mutiny: they moved away from the Reactive Streams API in favor of Java's Flow API.

Obviously, there was also the expected javax/jakarta fun.

Other than that, I tried to follow the upgrade principles outlined [here](https://github.com/quarkiverse/quarkiverse/wiki/Migrating-to-Quarkus-3.x). 

I tested locally including native mode, all tests pass (using GraalVM 22.3.1 + java 11).

Here is what I would suggest:

* Create a 1.1.x branch for future updates targeting Quarkus 2.x before merging this PR;
* Merge (well, rebase actually) this PR into main branch when approved;
* Release 1.2.0-alpha1 as soon as possible so that people can start integrating and testing.

Regarding the version number: the changes in this PR are NOT backwards compatible because 1) Quarkus 3 is not backwards compatible with Quarkus 2 and 2) The Mutiny changes are not backwards compatible either. This should warrant a major version bump, BUT: for most users the upgrade should be a no-brainer. With that in mind, I decided that a minor bump instead was "good enough", so I bumped to 1.2.0-alpha1-SNAPSHOT (we did the same when we switched from Quarkus 1 to Quarkus 2). But I'm happy to change this to a major bump instead.

\cc @ppalaga


ppalaga:
Thanks a lot @adutra! 
I [tested](https://github.com/apache/camel-quarkus/pull/4612) Camel Quarkus against this PR and it works well. 

> Regarding the version number: the changes in this PR are NOT backwards compatible because 1) Quarkus 3 is not backwards compatible with Quarkus 2 and 2) The Mutiny changes are not backwards compatible either. This should warrant a major version bump, BUT: for most users the upgrade should be a no-brainer. With that in mind, I decided that a minor bump instead was "good enough", so I bumped to 1.2.0-alpha1-SNAPSHOT (we did the same when we switched from Quarkus 1 to Quarkus 2). But I'm happy to change this to a major bump instead.

I'd vote for bumping the major, because other projects in the ecosystem do it like that. It might be easier for end users to see which versions of cassandra-quarkus are compatible with which Quarkus major versions. My 2 cents. I'll be also fine with 1.2.0-alpha1.

adutra:
Thanks for the feedback @ppalaga ! 

@absurdfarce or @msmygit could somebody please take a look at this and release an alpha version soon-ish? I'll let you decide whether this should be 1.2.0 or 2.0.0 :-)

Thanks 🙏

msmygit:
Since the Beta1 was released a couple days ago, I submitted a PR here, https://github.com/adutra/cassandra-quarkus/pull/1 to get it upgraded to that along with this.

ppalaga:
@absurdfarce Interesting! Thanks for the information.

absurdfarce:
You're quite welcome @ppalaga!

There will probably be other notifications elsewhere but I'll mention it here since it's directly relevant: the release was finished up last night.  1.2.0-alpha1 should now be available via Maven Central.  If you have any problems with it let me know!

@adutra you might be interested in the above as well. :)

adutra:
Thanks @absurdfarce, we already integrated it in our quarkus3 branch :-)

absurdfarce:
Hey @gsmet, regarding your question about Cassandra Quarkus and the Quarkus Platform... I think the schedule you outline will work for us.  1.2.0-alpha1 is now out on Maven and we've got [another PR](https://github.com/datastax/cassandra-quarkus/pull/227) up for CR1.  I suspect we should be able to get a final 1.2.0 release out between the release of 3.0 final on ~4/12 and the platform announcement on ~4/26.

Issue #225, opened 2023-03-24
CassandraHealthCheckIT fails when run on M1 Macs


absurdfarce:
Test is unable to run the underlying Docker image:

```
[INFO] -------------------------------------------------------                                                                                                                                                        
[INFO]  T E S T S                                                                                                                                                                                                     
[INFO] -------------------------------------------------------                                                                                                                                                        
[INFO] Running com.datastax.oss.quarkus.tests.CassandraHealthCheckIT                                                                                                                                                  
2023-03-24 15:56:08,974 WARN  [io.qua.res.mut.dep.ResteasyMutinyProcessor] (build-93) The quarkus-resteasy-mutiny extension is deprecated. Switch to RESTEasy Reactive instead.                                       
This extension adds support for Uni and Multi to RESTEasy Classic, without using the reactive execution model, as RESTEasy Classic does not use it. To properly integrate Mutiny and RESTEasy, use RESTEasy Reactive. 
See https://quarkus.io/guides/getting-started-reactive for detailed instructions                                                                                                                                      
2023-03-24 15:56:10,612 INFO  [🐳 .3.4]] (pool-3-thread-1) Creating container for image: testcontainers/ryuk:0.3.4                                                                                                    
2023-03-24 15:56:10,739 INFO  [🐳 .3.4]] (pool-3-thread-1) Container testcontainers/ryuk:0.3.4 is starting: 7161b3a855b969f97f658e8241d8f08ecbe62e2eb421df2484b2d57a1ede46a2                                          
2023-03-24 15:56:11,181 INFO  [🐳 .3.4]] (pool-3-thread-1) Container testcontainers/ryuk:0.3.4 started in PT0.584676S                                                                                                 
2023-03-24 15:56:11,188 INFO  [com.dat.oss.qua.tes.CassandraTestResource] (pool-3-thread-1) Container datastax/dse-server:6.8.25 starting...  
...
2023-03-24 15:56:16,690 ERROR [🐳 .8.25]] (pool-3-thread-1) Log output from the failed container:                                                                                                                     
exec /entrypoint.sh: exec format error                                                                                                                                                                                
                                                                                                                                                                                                                      
[ERROR] Tests run: 1, Failures: 0, Errors: 1, Skipped: 0, Time elapsed: 9.345 s <<< FAILURE! - in com.datastax.oss.quarkus.tests.CassandraHealthCheckIT                                                               
[...]

PR #226, opened 2023-03-29
Skip DSE integration tests if we're not on amd64

absurdfarce:
Fix for https://github.com/datastax/cassandra-quarkus/issues/225

### DIFF
--- a/test-framework/src/main/java/com/datastax/oss/quarkus/test/SkipIfNotAmd64.java
+++ b/test-framework/src/main/java/com/datastax/oss/quarkus/test/SkipIfNotAmd64.java
@@ -1,0 +1,10 @@
+package com.datastax.oss.quarkus.test;
+
+import org.junit.jupiter.api.extension.ExtendWith;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(CheckAmd64Condition.class)
+public @interface SkipIfNotAmd64 { }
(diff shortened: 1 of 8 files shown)


absurdfarce:
@adutra For some reason I can't add you as a reviewer but if you have the cycles I'd love for you to take a pass at this as well

msmygit (approved):
LGTM 👍🏼

test-framework/src/main/java/com/datastax/oss/quarkus/test/SkipIfNotAmd64.java
```diff
@@ -0,0 +1,24 @@
...
+ * limitations under the License.
+ */
+package com.datastax.oss.quarkus.test;
+
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import org.junit.jupiter.api.extension.ExtendWith;
+
+@Retention(RetentionPolicy.RUNTIME)
+@ExtendWith(CheckAmd64Condition.class)
```
adutra, line 23: You could annotate this with `@java.lang.annotation.Inherited` and then use the annotation directly on `DseTestBase` instead of using it on each concrete subclass.

adutra (approved):
Very elegant solution, LGTM.

test-framework/src/main/java/com/datastax/oss/quarkus/test/SkipIfNotAmd64.java
absurdfarce: Excellent point, I like that solution quite a bit better!  Changing now...

absurdfarce: Bah, my system clock was off (bad VM!) so the commit actually shows in history _before_ @adutra made his comment mentioning it.  [Relevant commit](https://github.com/datastax/cassandra-quarkus/pull/226/commits/e915f273bf96223fd1501919be240ee431300313) is above, tests for Java11 and Java17 running now.

PR #229, opened 2023-04-13
Upgrade to Quarkus 3.0.0.Final

absurdfarce:
Updated release version to 1.2.0 now that we're at a final release

### DIFF
--- a/deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
+++ b/deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
@@ -64,73 +64,90 @@ class CassandraClientProcessor {
 
   @BuildStep
   List<ReflectiveClassBuildItem> registerGraphForReflection() {
+    ReflectiveClassBuildItem.Builder builder =
+        ReflectiveClassBuildItem.builder(
+                "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal")
+            .constructors(true)
+            .methods(true)
+            .fields(true);
     return Arrays.asList(
         // Required for the driver DependencyCheck mechanism
-        new ReflectiveClassBuildItem(
-            true, true, "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal"),
+        builder.build(),
         // Should be initialized at build time:
-        new ReflectiveClassBuildItem(
-            true, true, "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerIoRegistryV3d0"),
-        new ReflectiveClassBuildItem(
-            true, true, "org.apache.tinkerpop.shaded.jackson.databind.deser.std.StdDeserializer"),
+        builder
+            .className(
+                new String[] {
+                  "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerIoRegistryV3d0"
+                })
+            .build(),
+        builder
+            .className(
+                new String[] {
+                  "org.apache.tinkerpop.shaded.jackson.databind.deser.std.StdDeserializer"
+                })
+            .build(),
         // Required by Tinkerpop:
         // TODO check if this is really all that is instantiated by reflection
-        new ReflectiveClassBuildItem(true, true, "org.apache.tinkerpop.gremlin.structure.Graph"),
-        new ReflectiveClassBuildItem(
-            true, true, "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerGraph"),
-        new ReflectiveClassBuildItem(
-            true, true, "org.apache.tinkerpop.gremlin.structure.util.empty.EmptyGraph"),
-        new ReflectiveClassBuildItem(
-            true,
-            true,
-            "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource"));
+        builder.className(new String[] {"org.apache.tinkerpop.gremlin.structure.Graph"}).build(),
+        builder
+            .className(
+                new String[] {"org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerGraph"})
+            .build(),
+        builder
+            .className(
+                new String[] {"org.apache.tinkerpop.gremlin.structure.util.empty.EmptyGraph"})
+            .build(),
+        builder
+            .className(
+                new String[] {
+                  "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource"
+                })
+            .build());
   }
 
   @BuildStep
   ReflectiveClassBuildItem registerGeometryForReflection() {
     // Required for the driver DependencyCheck mechanism
-    return new ReflectiveClassBuildItem(
-        false, false, false, false, "com.esri.core.geometry.ogc.OGCGeometry");
+    return ReflectiveClassBuildItem.builder("com.esri.core.geometry.ogc.OGCGeometry").build();
   }
 
   @BuildStep
   List<ReflectiveClassBuildItem> registerJsonForReflection() {
     // Required for the driver DependencyCheck mechanism
+    ReflectiveClassBuildItem.Builder builder =
+        ReflectiveClassBuildItem.builder("com.fasterxml.jackson.core.JsonParser");
     return Arrays.asList(
-        new ReflectiveClassBuildItem(
-            false, false, false, false, "com.fasterxml.jackson.core.JsonParser"),
-        new ReflectiveClassBuildItem(
-            false, false, false, false, "com.fasterxml.jackson.databind.ObjectMapper"));
+        builder.build(),
+        builder.className(new String[] {"com.fasterxml.jackson.databind.ObjectMapper"}).build());
   }
 
   @BuildStep
   List<ReflectiveClassBuildItem> registerReactiveForReflection() {
     // Required for the driver DependencyCheck mechanism
     return Collections.singletonList(
-        new ReflectiveClassBuildItem(false, false, false, false, "org.reactivestreams.Publisher"));
+        ReflectiveClassBuildItem.builder("org.reactivestreams.Publisher").build());
   }
 
   @BuildStep
   List<ReflectiveClassBuildItem> registerLz4ForReflection(
       CassandraClientBuildTimeConfig buildTimeConfig) {
     if (buildTimeConfig.protocolCompression.equalsIgnoreCase("lz4")) {
+      ReflectiveClassBuildItem.Builder builder =
+          ReflectiveClassBuildItem.builder("net.jpountz.lz4.LZ4Compressor")
+              .constructors(true)
+              .fields(true);
       return Arrays.asList(
-          new ReflectiveClassBuildItem(true, false, true, "net.jpountz.lz4.LZ4Compressor"),
-          new ReflectiveClassBuildItem(true, false, true, "net.jpountz.lz4.LZ4JavaSafeCompressor"),
-          new ReflectiveClassBuildItem(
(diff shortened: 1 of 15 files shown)


deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
```diff
@@ -64,73 +64,90 @@ FeatureBuildItem feature() {
 
   @BuildStep
   List<ReflectiveClassBuildItem> registerGraphForReflection() {
+    ReflectiveClassBuildItem.Builder builder =
+        ReflectiveClassBuildItem.builder(
+                "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal")
+            .constructors(true)
+            .methods(true)
+            .fields(true);
```
absurdfarce, line 72: @msmygit (and @adutra if you have the cycles) this is the part of the upgrade I'm most concerned about.  I was seeing a lot of warnings about the deprecated constructors of ReflectiveClassBuildItem so it seemed worthwhile to shift over to the new builder usage.  Unfortunately I'm not completely paranoid that I might've accidentally dropped a class in translation so if either (or both) of you have cycles to give this a good once-over and make sure all the expected classes are still covered I'd greatly appreciate it.

msmygit: LGTM. Let's also get @adutra's eyes on it. 👍🏼

msmygit (approved):

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
adutra: Why do we need `.fields(true)` here and not elsewhere? The original code had `true, true` for all items in this method.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
```diff
@@ -64,73 +64,90 @@ FeatureBuildItem feature() {
...
+        builder.build(),
         // Should be initialized at build time:
-        new ReflectiveClassBuildItem(
-            true, true, "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerIoRegistryV3d0"),
-        new ReflectiveClassBuildItem(
-            true, true, "org.apache.tinkerpop.shaded.jackson.databind.deser.std.StdDeserializer"),
+        builder
+            .className(
+                new String[] {
+                  "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerIoRegistryV3d0"
```
adutra, line 80: Can we pass more than one class name here? In this case we might be able to create one builder for all the class names to create items for? (I haven't actually tried it – would need to check the javadocs to understand why this method takes a `String[]` – which is surprising.)

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
```diff
@@ -64,73 +64,90 @@ FeatureBuildItem feature() {
...
         // Required for the driver DependencyCheck mechanism
-        new ReflectiveClassBuildItem(
-            true, true, "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal"),
+        builder.build(),
         // Should be initialized at build time:
-        new ReflectiveClassBuildItem(
-            true, true, "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerIoRegistryV3d0"),
-        new ReflectiveClassBuildItem(
-            true, true, "org.apache.tinkerpop.shaded.jackson.databind.deser.std.StdDeserializer"),
+        builder
```
adutra, line 77: Aren't we reusing the builder instance after a call to `build()`? This looks suspicious to me (not exactly the typical builder pattern). At least we'd need to be sure that the builder isn't accumulating data from before the last `build()` invocation.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
absurdfarce: Good eye @adutra!  The (boolean, boolean, String) [constructor](https://github.com/quarkusio/quarkus/blob/main/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ReflectiveClassBuildItem.java#L66-L69) is for (methods, fields, classes).  So we _do_ need `.fields(true)` here... but `.constructors(true)` is not required.

I'll remove it.

absurdfarce: Done

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
absurdfarce: Yeah, after reviewing this I realize I have this entirely wrong.  Somehow I misread the [underlying source](https://github.com/quarkusio/quarkus/blob/main/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ReflectiveClassBuildItem.java) to say that a single class name could be passed via `ReflectiveClassBuilderItem.builder()` but that String arrays were required for the builder methods.  Basically I missed the vararg in the static builder() method, and what's worse I remember thinking that that mismatch made for a very weird API (_and_ I remembering why the Quarkus guys would do such a thing).. but apparently I never double-checked that.

Anyway, I'll re-work all of this to be more sane... most of what I have here is outright wrong.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
```diff
@@ -206,26 +197,30 @@ void setupSslSupport(
...
-        return Arrays.asList(
-            new ReflectiveClassBuildItem(
-                true,
-                true,
-                "com.datastax.oss.driver.internal.metrics.microprofile.MicroProfileMetricsFactory"),
-            new ReflectiveClassBuildItem(true, true, TaggingMetricIdGenerator.class));
+        clzStream =
+            Stream.of(
+                TaggingMetricIdGenerator.class.getName(),
+                "com.datastax.oss.driver.internal.metrics.microprofile.MicroProfileMetricsFactory");
```
absurdfarce, line 214: It sure _seems_ like we need an additional else case here to handle the situation in which neither if block above is invoked.  That would correlate to metrics of some kind being in play but it being neither Micrometer or MicroProfile.  Is there a guard somewhere else that I missed which prevents this case from coming up?

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
```diff
@@ -64,76 +64,67 @@ FeatureBuildItem feature() {
...
+            // Required by Tinkerpop:
+            // TODO check if this is really all that is instantiated by reflection
+            "org.apache.tinkerpop.gremlin.structure.Graph",
+            "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerGraph",
+            "org.apache.tinkerpop.gremlin.structure.util.empty.EmptyGraph",
+            "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource")
+        .map(
+            (clzName) ->
+                ReflectiveClassBuildItem.builder(clzName).methods(true).fields(true).build())
+        .collect(Collectors.toList());
```
absurdfarce, line 83: This implementation deliberately mirrors the previous approach of creating a distinct RCBI instance for every class name we wish to account for.  The [relevant RCBI impl](https://github.com/quarkusio/quarkus/blob/3.0.0.Final/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ReflectiveClassBuildItem.java) seems to suggest that we _might_ be able to accomplish the same thing with a single RCBI instance fed with multiple class names.  In the interest of expediency I'm not pursuing that path at the moment... that's what the TODO above is meant to remind me about.

absurdfarce:
Okay @adutra, this has been completely re-worked into something that should be _much_ saner.  When you have the time would you mind taking another pass?  Same thing as before; I really want a second set of eyes to confirm I didn't get any of the field/constructor/method params mucked up.

That said, if you see me doing anything _else_ dumb with the Java language please feel free to flag that too. :(

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
```diff
@@ -64,76 +64,67 @@ FeatureBuildItem feature() {
...
-            true, true, "org.apache.tinkerpop.gremlin.structure.util.empty.EmptyGraph"),
-        new ReflectiveClassBuildItem(
-            true,
-            true,
-            "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource"));
+
+    /* TODO: We're creating an RCBI instance per class here to match up with the pre-Quarkus3.0 usage.
+     * Check to see if we can just simplify this to something like RCBI.builder(names).foo().build() */
+    return Stream.of(
+            "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal",
```
adutra, line 71:
```suggestion
            // Required for the driver DependencyCheck mechanism
            "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal",
            // Should be initialized at build time:
```

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
adutra: Looking at the javadocs of `metricsSupported`:

```
Returns:
true if this factory supports the named metrics system. Arbitrary strings are allowed. Constants are present for a few.
See Also:
io.quarkus.runtime.metrics.MetricsFactory.MICROMETER, io.quarkus.runtime.metrics.MetricsFactory.MP_METRICS
```

So while it seems that Micrometer and MicroProfile are the only ones officially supported, other metrics providers are in theory accepted.

We should indeed add an `else` block and maybe log a warning that driver metrics won't work for any other provider than the two already handled.

adutra (changes requested):

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
absurdfarce: Done.  It seemed simplest to just log a message indicating which metric impls were supported.  I went back and forth on the level but this seemed like something worthy of a warn.

msmygit (approved):
LGTM

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
msmygit: @absurdfarce I _believe_ `.constructors(true)` is indeed required because the other constructors `(methods,fields,classes)` is actually setting the `constructors` to `true` anyways [here](https://github.com/quarkusio/quarkus/blob/3.0.0.Final/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ReflectiveClassBuildItem.java#L84-L96). WDYT?

absurdfarce:
In the interest of time I'm going to merge this PR and finish up the release today.  If there are things we need to go back and revisit afterwards we can do that in the next version.

PR #232, opened 2023-04-14
Refactor the way we use RCBI

msmygit:
Refactor the way we use RCBI

### DIFF
--- a/deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
+++ b/deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
@@ -215,7 +215,7 @@ class CassandraClientProcessor {
                 TaggingMetricIdGenerator.class.getName(),
                 "com.datastax.oss.driver.internal.metrics.microprofile.MicroProfileMetricsFactory");
       } else {
-    	  LOG.warn("The cassandra-quarkus plugin only supports Micrometer or MicroProfile metrics");
+        LOG.warn("The cassandra-quarkus plugin only supports Micrometer or MicroProfile metrics");
       }
       return clzStream
           .map(
--- a/deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
+++ b/deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
@@ -65,22 +65,23 @@ class CassandraClientProcessor {
   @BuildStep
   List<ReflectiveClassBuildItem> registerGraphForReflection() {
 
-    /* TODO: We're creating an RCBI instance per class here to match up with the pre-Quarkus3.0 usage.
-     * Check to see if we can just simplify this to something like RCBI.builder(names).foo().build() */
-    return Stream.of(
-            "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal",
-            "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerIoRegistryV3d0",
-            "org.apache.tinkerpop.shaded.jackson.databind.deser.std.StdDeserializer",
-            // Required by Tinkerpop:
-            // TODO check if this is really all that is instantiated by reflection
-            "org.apache.tinkerpop.gremlin.structure.Graph",
-            "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerGraph",
-            "org.apache.tinkerpop.gremlin.structure.util.empty.EmptyGraph",
-            "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource")
-        .map(
-            (clzName) ->
-                ReflectiveClassBuildItem.builder(clzName).methods(true).fields(true).build())
-        .collect(Collectors.toList());
+    return Collections.singletonList(
+        ReflectiveClassBuildItem.builder(
+                // Required for the driver DependencyCheck mechanism
+                "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversal",
+                // Should be initialized at build time:
+                "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerIoRegistryV3d0",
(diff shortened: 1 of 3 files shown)


absurdfarce:
So I think we probably _do_ want to move to something like this in the near future.  That idea is really what [this comment](https://github.com/datastax/cassandra-quarkus/pull/229#discussion_r1166087933) was all about.  I'm just reluctant to make this change right before a 3.0.0.Final release; I'd prefer to do it in the _next_ release after, presumably when we've had some time to look into the implications (if any) and better understand what consequences (if any) this change might have.

Make sense?

So I guess I'm saying let's leave this around for now but merge it _after_ 1.2.0 goes out.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
```diff
@@ -65,24 +65,23 @@ FeatureBuildItem feature() {
...
+                // Should be initialized at build time:
+                "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerIoRegistryV3d0",
+                "org.apache.tinkerpop.shaded.jackson.databind.deser.std.StdDeserializer",
+                // Required by Tinkerpop:
+                // TODO check if this is really all that is instantiated by reflection
+                "org.apache.tinkerpop.gremlin.structure.Graph",
+                "org.apache.tinkerpop.gremlin.tinkergraph.structure.TinkerGraph",
+                "org.apache.tinkerpop.gremlin.structure.util.empty.EmptyGraph",
+                "org.apache.tinkerpop.gremlin.process.traversal.dsl.graph.GraphTraversalSource")
+            .constructors(true)
```
absurdfarce, line 81: Looks like constructors(true) was added here when it wasn't present in the original

pom.xml
```diff
@@ -33,7 +33,7 @@
     Note: when upgrading the Quarkus version or the DataStax Java driver version, make sure that you
     upgrade them in the quickstart module too.
     -->
-    <quarkus.version>3.0.0.Final</quarkus.version>
+    <quarkus.version>3.1.0.CR1</quarkus.version>
```
absurdfarce, line 36: Let's revert this change; I'd like to keep this PR focused on the RCBI changes only.

quickstart/pom.xml
```diff
@@ -27,7 +27,7 @@
   <inceptionYear>2020</inceptionYear>
   <properties>
     <java.version>11</java.version>
-    <quarkus.version>3.0.0.Final</quarkus.version>
+    <quarkus.version>3.1.0.CR1</quarkus.version>
```
absurdfarce, line 30: Please revert this one as well... same rationale as above.

absurdfarce (changes requested):
Looks pretty sane over all, just a couple of minor changes and I think we'll be all set.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
msmygit: Because previously constructors were true by default and that [method is now deprecated](https://github.com/quarkusio/quarkus/blob/3.0.3.Final/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ReflectiveClassBuildItem.java#L80-L87) and hence the change.

msmygit: Before you ask, even in [2.x Quarkus version, constructors were always `true`](https://github.com/quarkusio/quarkus/blob/2.16.7.Final/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ReflectiveClassBuildItem.java#L24-L26).

absurdfarce: The specific constructor you cite for 3.0 does calls another constructor in turn and it does set the "constructors" param to true when it does so.  But that's only because the constructor in question doesn't accept a "constructors" argument.  [Other](https://github.com/quarkusio/quarkus/blob/3.0.3.Final/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ReflectiveClassBuildItem.java#L71-L78) [constructors](https://github.com/quarkusio/quarkus/blob/3.0.3.Final/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ReflectiveClassBuildItem.java#L89-L96) do accept such an argument.

It's true that these methods are deprecated, but this appears to be the case for all RCBI constructors; based on the comments it's clear they're trying to steer developers to use the builders instead (which is entirely sensible).  The builder still very much includes a "constructors" [method](https://github.com/quarkusio/quarkus/blob/3.0.3.Final/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ReflectiveClassBuildItem.java#L196-L199) which accepts a boolean argument but [the default](https://github.com/quarkusio/quarkus/blob/3.0.3.Final/core/deployment/src/main/java/io/quarkus/deployment/builditem/nativeimage/ReflectiveClassBuildItem.java#L181) is still "true".

So at best including "constructors(true)" here is redundant (since it's just restating the default).

With respect to Quarkus 2.x: the set of constructors there look very similar (they just aren't deprecated yet).  As discussed above some of the constructors set "constructors" to true if they don't accept that as an argument but there are plenty of constructors in 2.x which do accept a boolean arg corresponding to "constructors"... so it very much comes down to which constructor a developer wants to make use of.

I guess I'd argue against changing the builder methods called here unless we have a clear benefit from it.  Seem sensible?

msmygit: Good point. Removed calling the `.constructors(true)` from here.

absurdfarce:
I just noticed: we need to change this PR around to go against main.  quarkus-300Final has been [merged into main](https://github.com/datastax/cassandra-quarkus/commit/43ad2c06bf44a4d1bb0e5448db75b88b2688d828) already.  In fact we probably ought to just delete the quarkus-300Final branch outright.

msmygit:
Closing this in favor of #236

PR #242, opened 2023-08-18
update quickstart with lombok and enum example

Pashnov:
update some dependencies to actual state with properties as it present in latest quarkus builds
add lombok library
add config to properly work with lombok
add enum type as example how to work with custom types add code how to add custom TypeCodec
update README.adoc
update fruits.html

### DIFF
--- a/quickstart/README.adoc
+++ b/quickstart/README.adoc
@@ -76,6 +76,7 @@ First, let's create our data model – represented by the `Fruit` class – as f
 
 [source,java]
 ----
+@Data
 @Entity
 @PropertyStrategy(mutable = false)
 public class Fruit {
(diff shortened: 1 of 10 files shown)


Pashnov:
I would like to push above changes to quickstart project, as that information is not present in internet (ot it's too hard to find) and, I think, it's useful who would like to start to use this library.

I've added type codec registration via `AppLifecycleBeanCodecRegistration` with higher priority that `CassandraClientStarter` run after with proper typeCodec and don't fail. I'm not sure that it's proper way but I haven't found other way how to make it work.

please review and if it's ok, please merge or let me know what should I fix.

cc: @gsmet @msmygit @adutra @tomekl007

PR #274, opened 2026-07-20
provide hook for additional customizations

### LINKED ISSUE #273: Provide customization hook to register custom codecs before CQLSession is started
When implementing custom codecs for user defined types there is currently no hook to properly integrate them in quarkus-cassandra in a way that the registry is up-to-date before cassandra-quarkus starts the CQLSession.

There is only a workaround which relies on internals by replacing the `QuarkusCqlSession` bean with an `Alternative`

```java
@ApplicationScoped
@Alternative
@Priority(1)
public class LazyvalCassandraCodecRegistrar {
  private final CassandraClientProducer delegate;

  @Inject
  public LazyvalCassandraCodecRegistrar(CassandraClientProducer delegate) {
    this.delegate = delegate;
  }

  @Produces
  @ApplicationScoped
  @Unremovable
  public CompletionStage<QuarkusCqlSession> produceCodecAwareSessionStage(
      CassandraClientConfig config, @MainEventLoopGroup EventLoopGroup mainEventLoop) {
    CompletionStage<QuarkusCqlSession> stage = delegate.produceQuarkusCqlSessionStage(config, mainEventLoop);
    return stage.thenApply(session -> {
        MutableCodecRegistry registry = (MutableCodecRegistry) session.getContext().getCodecRegistry();
        registry.register(new LazyvalCassandraCodecs.QuantityCodec());
        registry.register(new LazyvalCassandraCodecs.EMailCodec());
        registry.register(new LazyvalCassandraCodecs.IsbnCodec());
        registry.register(new LazyvalCassandraCodecs.OrderDateCodec());
        return session;
    });
  }
}
```

This is not reliable and might fail when changing versions.

A customization singleton like for Jackson/JSONB would be nice.

lostiniceland:
resolves #273 

Provides an extension mechanism to modify the CqlSession before it is created.

```java
@Singleton
public class CassandraCodecCustomizer implements CqlSessionCustomizer {
    @Override
    public void customize(QuarkusCqlSessionBuilder quarkusCqlSessionBuilder) {
        quarkusCqlSessionBuilder.addTypeCodecs(new MyCustomCodec());
    }
}
```
The approach is a mere copy of Quarkus ObjectMapperCustomizer or JsonbConfigCustomizer and the appropriate pattern as discussed in the [Quarkus chat](https://quarkusio.zulipchat.com/#narrow/channel/187030-users/topic/.E2.9C.94.20Cassandra-Quarkus.3A.20feedback.20needed.20on.20extension.20mechanism/with/611824025).

I am not sure if this is a good idea to pass the Builder to the customizers as user can wreak havoc. To mitigate at least a bit, the customizers are called early so any important overrides are done later.


lostiniceland:
@absurdfarce @adutra can you approve the workflow please, so I can see if anything is missing. local builds where successful.

And a question regarding what should be exposed by the interface: instead of `QuarkusCqlSessionBuilder` we could pass it just as `SessionBuilder` which reduces the API surface.

adutra:
@lostiniceland I'm not a maintainer of this project anymore.

lostiniceland:
Sorry, I just had a look at you two and saw you are somewhat involved in the Cassandra space.

absurdfarce:
@lostiniceland I've approved your workflow to run.  I'll also be reviewing your PR and working to get it merged.  Unfortunately I'm going to be out of the office until early August but I'll revisit the state of things when I'm back in.

In general, while I don't disagree that your approach seems reasonable I'm worried about introducing a totally new configuration mechanism that isn't used anywhere else in the extension.  I'm just wondering; have you looked at [CassandraClientProcessor](https://github.com/datastax/cassandra-quarkus/blob/1.4.1/deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java) (and by extension [CassandraClientRecorder](https://github.com/datastax/cassandra-quarkus/blob/1.4.1/runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientRecorder.java)) with the intent of seeing if there's a way to provide a build-time config which could be leveraged to accomplish the same thing?  Seems like we could add something like a TypeCodecProvider (or something in that vein) which could be added at build-time and subsequently modify generated QuarkusCqlSession instances along the same lines as what's already done in CassandraClientProcessor.

Thoughts?

lostiniceland:
I just haven't been able to get additional codecs registered without using the "hack" in the linked issue. I will take a look at your suggestions as build-time config is usually better, but following the Quarkus customization-pattern was the easiest solution to get it working. As I've modified the integration-tests it should be easier to test other solutions now.

Edit: had a quick look at the Processor and I agree that this would be the better approach. It also would allow me to reduce complexity in my downstream project where I need to apply codecs

lostiniceland:
I've changed the implementation (with the help of Claude) to use Jandex to look up TypeCodecs.

The lookup was made smart enough to also not just find public classes but also factory-methods which provide package-private or static inner classes.

I've kept the previous commit, but will rebase once a decision has been made which solution to keep.

lostiniceland:
This is an example why I would like to make use of a "discovery" factory-method as. My annotation processor generates code which hides custom codecs as package-private static classes in order no to pollute the filesystem and encapsulate the code as much as possible as it is mostly irrelevant to application developers.


```
@Generated("com.qualityminds.lazyval.processor.internal.codegen.cassandra.CassandraCodecGenerator")
public final class LazyvalCassandraCodecs {
  private LazyvalCassandraCodecs() {
  }

  /**
   * Returns an array of all generated {@link TypeCodec}s for lazyval wrapper types.
   */
  public static TypeCodec<?>[] all() {
    return new TypeCodec[] {
            new EMailCodec(),
            new IsbnCodec()
        };
  }


  static class EMailCodec extends MappingCodec<String, EMail> {
    ...
  }

  static class IsbnCodec extends MappingCodec<String, Isbn> {
    ...
  }

}
```

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
```diff
@@ -277,6 +296,162 @@ private boolean checkMicrometerMetricsFactoryPresent() {
...
+    return new AdditionalIndexedClassesBuildItem(
+        TypeCodec.class.getName(),
+        MappingCodec.class.getName(),
+        PrimitiveBooleanCodec.class.getName(),
+        PrimitiveByteCodec.class.getName(),
+        PrimitiveDoubleCodec.class.getName(),
+        PrimitiveFloatCodec.class.getName(),
+        PrimitiveIntCodec.class.getName(),
+        PrimitiveLongCodec.class.getName(),
+        PrimitiveShortCodec.class.getName());
```
absurdfarce: I'm assuming the primitive codecs are here because some of the MappingCodec impls here map to or from "primitive" equivalents like Integer.  The implication is that if a user wants to do something like that they have to include codecs for the mapped-from and mapped-to codecs in the Jandex index, correct?

If that is the case we might want to spell that out in docs somewhere... although I can potentially add that later.

It's also worth noting that the Java driver [already defines](https://github.com/apache/cassandra-java-driver/blob/4.19.3/core/src/main/java/com/datastax/oss/driver/internal/core/type/codec/registry/CodecRegistryConstants.java#L35-L59) a common set of TypeCodec instances for it's internal CodecRegistry.  It might be worthwhile just re-using that here rather than running through all the primitive codecs (and thus creating a space where we have the same info in two places).

integration-tests/main/src/test/java/com/datastax/oss/quarkus/tests/CustomerResourceIT.java
```diff
@@ -22,6 +22,7 @@
 
 import com.datastax.oss.quarkus.test.CassandraTestResource;
 import com.datastax.oss.quarkus.tests.entity.Address;
+import com.datastax.oss.quarkus.tests.entity.Born;
```
absurdfarce: Adding "Born" (with it's custom codec) here actually makes this test cover two different cases:

* The ability to create/delete instances (which it was already testing)
* The ability to handle custom codecs

It's probably worth decoupling these ideas into two distinct tests.  I want to make sure we have a single basic test that should work in the common case; that's a useful canary.  We clearly also need integration tests for user-supplied codecs but that should probably live elsewhere.

I'm happy to take on that task in the near future.  I'm going to be working on a cassandra-quarkus release soon and I'd very much like to get this feature in that release so I don't see any reason to hold up moving forward with this PR to address this testing situation.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
```diff
@@ -277,6 +296,162 @@ private boolean checkMicrometerMetricsFactoryPresent() {
...
+  @BuildStep
+  void discoverTypeCodecProviders(
+      CombinedIndexBuildItem combinedIndex,
+      BuildProducer<CassandraTypeCodecProviderBuildItem> typeCodecProviders) {
+    for (ClassInfo clz : combinedIndex.getIndex().getKnownClasses()) {
+      if (isDriverClass(clz.name().toString())) {
+        continue;
+      }
+      boolean declaringClassIsPublic = Modifier.isPublic(clz.flags());
+      for (MethodInfo method : clz.methods()) {
```
absurdfarce: If I'm following along correctly this will scan every method of every public class in the index to see if it returns a TypeCodec.  Is it worth adding an annotation (or some other mechanism) to allow public methods that return TypeCodecs to opt out from inclusion here?

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
```diff
@@ -277,6 +296,162 @@ private boolean checkMicrometerMetricsFactoryPresent() {
...
+   * Finds all {@link TypeCodec} implementations in the application and turns each one into a {@link
+   * CassandraTypeCodecBuildItem}.
+   *
+   * <p>Codecs are instantiated by {@link CassandraClientProducer} when the session is built, hence
+   * only public classes with a public no-arg constructor can be registered this way.
+   */
+  @BuildStep
+  void discoverTypeCodecs(
+      CombinedIndexBuildItem combinedIndex, BuildProducer<CassandraTypeCodecBuildItem> typeCodecs) {
+    for (ClassInfo codec : combinedIndex.getIndex().getAllKnownImplementations(TYPE_CODEC)) {
```
absurdfarce: Similar to the comment for type codec providers: should we have an annotation (or some other marker) which allows implementers to say not to include a TypeCodec in the list of codecs passed to the underlying Session?

I'm not immediately sure what such a use case might be so there answer might very well be "no"... but it seemed worthwhile to at least discuss.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientProducer.java
```diff
@@ -192,6 +208,70 @@ private void configureProtocolCompression(
     configLoaderBuilder.withString(DefaultDriverOption.PROTOCOL_COMPRESSION, protocolCompression);
   }
 
+  /**
+   * Registers the type codecs discovered at build time.
+   *
+   * <p>Codec classes are collected by the extension's build steps and instantiated here, rather
+   * than at build time, so that codec instances are never captured in a native image heap.
```
absurdfarce: I'm not sure I understand the difference here.  We're including the required TypeCodec classes in the native image anyway via the ReflectiveClassBuildItem used in CassandraClientProcessor.  The class _has_ to be included in the native image in order to support reflective instantiation at deploy time at all.

I did note earlier that the build items generated at deployment time all reference class/method names rather than Class instances.  There is part of me wondering if we wouldn't be better off doing the reflection at build-time and then just passing along the TypeCodec classes to add to the QuarkusCqlSession at runtime.

absurdfarce (changes requested):
Excellent work @lostiniceland!

I generally approve of the approach here and the use of jandex seems entirely reasonable to me.  There's one spot here where I think we can benefit from a Java driver built-in and a few questions I'd like to get your feedback on but overall this PR is pretty close as it stands.

integration-tests/main/src/test/java/com/datastax/oss/quarkus/tests/CustomerResourceIT.java
lostiniceland: Agreed. I will split this up.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
lostiniceland: I agree that some configuration option is required. 

For instance I forgot that there might be test-fixtures, which the user want to exclude. 
Furthermore, there is an issue when two codecs for the same type are detected (for whatever reason). Currently there is a debug message telling the user that a codec was already registered, but the problem is that the user has no control over which one wins because it is dependent an unordered list (first one wins).

Instead of adding another runtime-feature (annotation) I would propose a configuration option which tackles the failure mode "Jandex found something I didn't mean".
Quarkus uses something for Arc along the lines `quarkus.arc.exclude-types` which is what I would lean toward.

```
quarkus.cassandra.type-codecs.exclude=com.acme.legacy.**,com.acme.FooCodec
quarkus.cassandra.type-codecs.discovery.enabled=true
```

- exclusion-rule for certain types
- a global kill-switch in case the user wants to register themselfs (though as I've figured out this is currently very hard and would require something like the extension-hook I proposed earlier) 

I am not sure on the last one, due to the fact that it currently is not really possible to customize the session, but on the other hand it might be useful in cases where the discovery misbehaves.

We might also consider an **opt-in** approach (which might be valid when the kill-switch is on)

```
quarkus.cassandra.type-codecs=com.acme.SpecialCodec  
```

We also might shortening the config to `quarkus.cassandra.codecs...` (not `type-codecs`)

integration-tests/main/src/test/java/com/datastax/oss/quarkus/tests/CustomerResourceIT.java
lostiniceland: I've added a dedicated commit that reverts the original tests and adds a dedicated one.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
lostiniceland: the initial rationale was that I might be necessary to extend those Primitive*Codec types, but it turned out that this is not even allowed. 
The only thing a user would use for custom codecs is a TypeCodec or MappingCodec so those types are the only ones Jandex needs to care about.

There is one gap, which is when there is an abstract codec somewhere which is then subclassed. The index will not find it and thus fail with a RuntimeException.
I propose to make use of the opt-in config-option described below, which would give the user a way to "fix" things Jandex missed.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
lostiniceland: After checking the two lookups (public accessible custom codecs & provider method) again a better solution might be two-fold:

1. config-options as described above for opt-out and "fixing" for codec-discovery
2. annotation-based opt-in for a provider-method

As the user would have full-control over the provider-method, I would propose that this annotation disables the discovery from 1. 
The annotation has the benefit that we do not have to scan every public method which might be a problem in big codebases for such a small feature.

@absurdfarce In case you have no strong objections, I would like to go ahead with this approach to see how it feels from a user-perspective. Fine-tuning can still be done afterwards.

deployment/src/main/java/com/datastax/oss/quarkus/deployment/internal/CassandraClientProcessor.java
lostiniceland: see https://github.com/datastax/cassandra-quarkus/pull/274#discussion_r4025062055

lostiniceland:
@absurdfarce can you please approve the workflow again.

I am pretty happy with the current solution. I was able to resolve your findings by having several configuration options as well as a new annotation for the provider-method so we dont have to look at every method in the index.

Next to the code I would need you input on the precedence rules:

1. codecs configured via inclusion will take precedence over excluded ones (you might exclude a whole package but still include a single codec)
2. when discovery is disable via configuration inclusions will still be registered and the same is true for annotated provider-methods

The rationale behind this was that I wanted to give the user the option to opt-out of the magic but still use the extension for configuration which are under his control.

If you could tell me where I should place any additional documentation I will be happy to add this as well.

runtime/src/main/java/com/datastax/oss/quarkus/runtime/internal/quarkus/CassandraClientProducer.java
lostiniceland: This is a leftover from the AI and has been removed. When all is green we should be fine.

absurdfarce:
@lostiniceland Workflow approved and should be running.

I need to take another pass on your recent changes.  I'm working on trying to get a Java driver release out at the moment but once that's done revisiting the cassandra-quarkus tickets are high on my list!

